PR #72, opened 2024-12-05
Gen thrift files again

StrongestNumber9:
Libthrift was updated in #61
Fixes #67 -- this is very experimental and I have no idea if this is even remotely correct thing to do.
Pinging @Laukkala as well to look at this

My manual changes were done in commit [0e4d405](https://github.com/teragrep/zep_01-migrate/pull/72/commits/0e4d4059f5e89560ed24315424cc752478bed6e4) and the rest of the automatic changes are in [1c3401b](https://github.com/teragrep/zep_01-migrate/pull/72/commits/1c3401bf6412e684edb717732dc95d34b4871f02)

Thrift files are in https://github.com/teragrep/zep_01-migrate/tree/master/zeppelin-interpreter/src/main/thrift

PR #73, opened 2024-12-09
Nuke download utils

### LINKED ISSUE #70: Nuke DownloadUtils
It will most likely nuke zeppelin-interpreter-integration tests but maybe it can be avoided by using proper dependency management

StrongestNumber9:
Built on top of #72 so handle that first


StrongestNumber9:
Just as a note, I am confused why this still works as sparkHome is null and whatever things it downloaded have been cleaned up locally, nothing should exist.

I built it in a fresh rockylinux:9 container with the following commands

```
    6  dnf install python python-unversioned-command maven java-11-openjdk
    7  mvn clean package -Pintegration
```

and it built properly, including integration tests where it was used so I have no idea what is the purpose of that class

kortemik (approved):

PR #139, opened 2025-01-21
Relocate PR template file

ronja-ui:
Changed the template's location since it doesn't automatically work if it's not right under .github directory.


StrongestNumber9 (approved):
lgtm

PR #195, opened 2025-03-19
Changed the name of python binary 

Laukkala:
Python binary name changed from python -> python3 to reference the correct name when running PythonInterpreter. Changed all instances of the python.zeppelin property, whether it's as a default value in code or in a configuration file default. Also updated a test that was explicitly setting this property.


StrongestNumber9 (approved):
Builds ok, looks ok, lgtm

PR #204, opened 2025-04-01
Restrict configuration interpreter from setting previously unknown keys

### LINKED ISSUE #108: Restrict configuration interpreter from setting previously unknown keys
ConfInterpreter should only be able to set properties that already exist. Trying to add a key that doesn't exist should result in an error.

### LINKED ISSUE #288: content verification for %spark.conf content
**Description**
content verification for %spark.conf content. the run on the conf should apply verification rules that the contents are valid

**Use case or motivation behind the feature request**
users may enter invalid configuration and this should be checked.

**Related issues**

**Additional context**
implementation needs to be discussed as it may change interpreter inteface

Laukkala:
See pagure link in comments for original peer-review comments

Resolves #108 
Resolves #288 

ConfInterpreter can now only update already existing properties, and it will throw an Exception if a user tries to set a property with a previously unknown key.

New keys should be created from the Interpreters page by authorized users, after which they can be updated by a ConfInterpreter.

Updated existing tests to comply with the new behaviour.


Laukkala:
This PR originates from previous version control platform, PR#72

Nefarious46 (dismissed):
Good work, looks good to me.

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/ConfInterpreter.java
```diff
@@ -67,11 +67,18 @@ public InterpreterResult interpret(String st, InterpreterContext context)
       finalProperties.putAll(getProperties());
       Properties newProperties = new Properties();
       newProperties.load(new StringReader(st));
+      // Verify that every key to be added exists within the Interpreter's defined properties already.
+      for (String propertyName : newProperties.stringPropertyNames()){
+        if(finalProperties.getProperty(propertyName) == null){
```
eemhu: Could `containsKey` be used instead? Would make it a bit cleaner.

Laukkala: Changed in commit [058dd3d](https://github.com/teragrep/zep_01/pull/204/commits/058dd3d913b2b6a15310f5c510d088e744b7b7c0)

eemhu (approved):

PR #209, opened 2025-04-09
Issue 86 websocket session confusion

Laukkala:
## Description

Moved the emission of COLLABORATIVE_MODE_STATUS messages so that they are only called when the connected users have actually changed. This should remove the COLLABORATIVE_MODE_STATUS updates being sent to other notebooks.

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

### DIFF
--- a/zeppelin-server/src/main/java/com/teragrep/zep_01/socket/ConnectionManager.java
+++ b/zeppelin-server/src/main/java/com/teragrep/zep_01/socket/ConnectionManager.java
@@ -172,21 +172,21 @@ public class ConnectionManager {
     }
   }
 
+  // Sends a response message indicating whether or not the notebook in question has multiple users connected, and lists the usernames of the connected users
   private void checkCollaborativeStatus(String noteId, List<NotebookSocket> socketList) {
     if (!collaborativeModeEnable) {
       return;
     }
-    boolean collaborativeStatusNew = socketList.size() > 1;
-    if (collaborativeStatusNew) {
+    boolean collaborativeStatus = socketList.size() > 1;
+    if (collaborativeStatus) {
       Message message = new Message(Message.OP.COLLABORATIVE_MODE_STATUS);
-      message.put("status", collaborativeStatusNew);
-      if (collaborativeStatusNew) {
-        HashSet<String> userList = new HashSet<>();
-        for (NotebookSocket noteSocket : socketList) {
-          userList.add(noteSocket.getUser());
-        }
-        message.put("users", userList);
+      message.put("status", collaborativeStatus);
+      // Create a list of users for the response
+      HashSet<String> userList = new HashSet<>();
+      for (NotebookSocket noteSocket : socketList) {
+        userList.add(noteSocket.getUser());
       }
+      message.put("users", userList);
       broadcast(noteId, message);
     }
   }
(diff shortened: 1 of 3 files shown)


zeppelin-server/src/main/java/com/teragrep/zep_01/socket/ConnectionManager.java
```diff
@@ -168,43 +167,28 @@ public void removeConnectionFromAllNote(NotebookSocket socket) {
...
-
-    Message message = new Message(Message.OP.COLLABORATIVE_MODE_STATUS);
-    message.put("status", collaborativeStatusNew);
-    if (collaborativeStatusNew) {
-      HashSet<String> userList = new HashSet<>();
-      for (NotebookSocket noteSocket : socketList) {
-        userList.add(noteSocket.getUser());
+      Message message = new Message(Message.OP.COLLABORATIVE_MODE_STATUS);
+      message.put("status", collaborativeStatusNew);
+      if (collaborativeStatusNew) {
```
Tiihott: This if-statement condition is always true.

zeppelin-server/src/test/java/com/teragrep/zep_01/socket/NotebookServerTest.java
```diff
@@ -683,6 +685,132 @@ public void testNoteRevision() throws IOException {
     assertEquals(0, note.getParagraphCount());
   }
 
+  @Test
+  public void testCollaborativeModeStatus() {
+    Assertions.assertDoesNotThrow(()->{
```
Tiihott: It would make it easier to debug issues if the assertion only covered the methods that could throw exceptions. Same goes for the next test.

zeppelin-server/src/test/java/com/teragrep/zep_01/socket/NotebookServerTest.java
```diff
@@ -683,6 +685,132 @@ public void testNoteRevision() throws IOException {
...
+      }
+      // create note from sock1
+      notebookServer.onMessage(sock1,
+              new Message(OP.NEW_NOTE)
+                      .put("name", noteName)
+                      .put("defaultInterpreterId", defaultInterpreterId).toJson());
+
+      String noteId = "";
+      for (Note note : notebook.getAllNotes()) {
+        if (note.getName().equals(noteName)) {
```
Tiihott: `Assertions.assertEquals(note.getName(), noteName);` would also work so there wouldn't be a need for if-else statement.

Tiihott (changes requested):
Few issues but otherwise looks OK.

zeppelin-server/src/test/java/com/teragrep/zep_01/socket/NotebookServerTest.java
Laukkala: Moved the beginning of the assertion to the latest position where all necessary variables are available

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/ConnectionManager.java
Laukkala: Removed

zeppelin-server/src/test/java/com/teragrep/zep_01/socket/NotebookServerTest.java
Laukkala: Simplified this part to remove the unnecessary loop and if-else statement

Tiihott (approved):
Tests pass and changes look good. LGTM.

kortemik (approved):

PR #230, opened 2025-07-03
Issue 227 Remove stateful elements from server-side datatable communication

Laukkala:
## Description
Simplifies communication between AJS_01 and ZEP_01 regarding server-side datatables.
Previous implementation used angular object registries to both receive and send data to UI when using DPL datatables. The UI and Server were both listening to changes in stateful elements to trigger some actions.
These elements have been removed from the server-side, and now data is provided in a uniform way.

This version uses a different method to send data to UI, and has formatted some values to be the same as is expected by the UI's datatable implementation. 

### Changes
Removed AJAXResponse & Watcher
Removed DTAnchorReady & Watcher
Removed DTServerupdate & Watcher
Removed PageRefresh & Watcher
AJAXRequest is still in use.

Table data is sent to UI via PARAGRAPH_UPDATE_OUTPUT messages instead of ANGULAR_OBJECT_UPDATE AJAXResponse messages.
Pagination requests are received from UI via PARAGRAPH_UPDATE_RESULT messages instead of ANGULAR_OBJECT_UPDATED AJAXRequest messages. Internally still uses AngularObjectRegistry to access the correct instance of DTTableDatasetNG.
AJAXRequest angular objects are now additionally identified with their associated paragraphId. 

Adds a header object to the paginated table data for use on UI side.
Adds a ResultsTotal and ResultsFiltered fields to table data for use on UI side.
Adds a draw value to table data for use on UI side.

Added new JSONTABLE InterpreterResult type.

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality


spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTHeader.java
eemhu, line 1: Improvement idea: Add ctor for `schema` and change static methods to non-static `xml()` and `json()`

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTHeaderTest.java
```diff
@@ -76,5 +76,10 @@ public void testDTHeader() throws ParserConfigurationException, TransformerExcep
         String e = "<thead><tr><th>_time</th><th>_raw</th><th>index</th><th>sourcetype</th><th>host</th><th>source</th><th>partition</th><th>offset</th></tr></thead>";
         assertEquals(e, DTHeader.schemaToHeader(Schema));
     }
+    @Test
+    public void testJSONDTHeader() throws ParserConfigurationException, TransformerException {
+        String e = "{\"_time\":\"\",\"_raw\":\"\",\"index\":\"\",\"sourcetype\":\"\",\"host\":\"\",\"source\":\"\",\"partition\":\"\",\"offset\":\"\"}";
```
eemhu: should build the expected json object instead of using string

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNgTest.java
```diff
@@ -202,22 +210,50 @@ public void testAJAXResponse() {
...
-                        "\"ID\":0," +
-                        "\"datalength\":49" +
-                        "}"
-                , response.toString()
-        );
+
+        JsonObject headers = Json.createReader(new StringReader(DTHeader.schemaToJsonHeader(testSchema))).readObject();;
+            JsonObject response = DTTableDatasetNg.DTNetResponse(formated, headers, 1, datasetAsJSON.size(),formated.size());
+
+            assertEquals("" +
```
eemhu: same here, these should be built with the JsonObjectBuilder instead of strings

zeppelin-server/src/main/java/com/teragrep/zep_01/service/NotebookService.java
```diff
@@ -662,6 +662,11 @@ public void updateParagraph(String noteId,
     callback.onSuccess(p, context);
   }
 
+  public void updateParagraphResult(String noteId, String paragraphId, String interpreterGroupId,int draw, int start, int length, String searchString,ServiceContext context, ServiceCallback<AngularObject> callback) throws IOException {
+    // Make an AJAXRequest based on given information
+    updateAngularObject(noteId,paragraphId,interpreterGroupId,"AJAXRequest_"+paragraphId,"{\"draw\":"+draw+",\"start\":"+start+",\"length\":"+length+",\"search\":{\"value\":\""+searchString+"\",\"regex\":false}}",context,callback);
```
eemhu: also here with the string concat

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTHeaderTest.java
Laukkala: Fixed in commit [0658a37](https://github.com/teragrep/zep_01/pull/230/commits/0658a37eec5eac9e0057af8554f59a2044d37eeb)

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNgTest.java
Laukkala: Fixed in commit [0658a37](https://github.com/teragrep/zep_01/pull/230/commits/0658a37eec5eac9e0057af8554f59a2044d37eeb)

zeppelin-server/src/main/java/com/teragrep/zep_01/service/NotebookService.java
Laukkala: This required adding a dependency to jakarta.json to zeppelin-server, as ZEP_01 has many different modules each with their own dependencies, but I imagine that it shouldn't be a problem. 
Fixed in commit [e94b71d](https://github.com/teragrep/zep_01/pull/230/commits/e94b71dd43609a0513d658d130b0d908c70c9607)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTHeader.java
Laukkala: Implemented in commit [1689d0f](https://github.com/teragrep/zep_01/pull/230/commits/1689d0ff747daf6d47cb0c2fb0fb2d3b0ebbdf92)

Refactoring these methods and thei usages within ZEP_01 to be non-static didn't cause any issues within ZEP_01, but if there is any other code in any other component that references these static methods, this might be a breaking change.

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNgTest.java
```diff
@@ -212,25 +211,45 @@ public void testAJAXResponse() {
...
+                    .add("sourcetype","stream")
+                    .add("host","host")
+                    .add("source","input")
+                    .add("partition","0")
+                    .add("offset",0)
+                    .add("origin","test data")
+                    .build();
+            dataBuilder.add(rowJson);
+        }
+        JsonArray data = dataBuilder.build();
```
eemhu, line 237: assert that dataBuilder has correct amount of rows

Laukkala: Fixed in commit [3b5e49e](https://github.com/teragrep/zep_01/pull/230/commits/3b5e49e07fce4d8ece839fc1886bd60e61975038)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -72,19 +73,12 @@ public final class DTTableDatasetNg extends AbstractUserInterfaceElement {
     private final Lock lock = new ReentrantLock();
 
     private final AngularObject<String> AJAXRequestAngularObject;
-    private final AngularObject<String> AJAXResponseAngularObject;
-    private final AngularObject<Boolean> DTAnchorReadyAngularObject;
-    private final AngularObject<Boolean> pageRefreshAngularObject;
-    private final AngularObject<Integer> DTServerUpdateAngularObject;
 
     private List<String> datasetAsJSON = null;
```
eemhu, line 74: Should make an issue about making these objects immutable.

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -174,8 +118,12 @@ void refreshPage() {
     void handeAJAXRequest(AJAXRequest ajaxRequest) {
         try {
             lock.lock();
-
-            updatePage(ajaxRequest);
+            // Apply defaults if parameters are not provided
+            int start = (ajaxRequest.getStart() == null) ? 0 : ajaxRequest.getStart();
+            int length = (ajaxRequest.getLength() == null) ? 25 : ajaxRequest.getLength();
```
eemhu: Same here, new issue about removing nulls from AjaxRequest.

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -249,40 +173,54 @@ public void setParagraphDataset(Dataset<Row> rowDataset) {
...
         }
+        try {
+            JsonObject response = SearchAndPaginate(draw, start,length,searchString);
+            String outputContent = "%jsontable\n" +
+                    response.toString();
+            getInterpreterContext().out().clear();
+            getInterpreterContext().out().write(outputContent);
+            getInterpreterContext().out().flush();
+        } catch (java.io.IOException e) {
+            LOGGER.error(e.toString());
```
eemhu, line 176: Could these errors be handled? Perhaps a new issue also?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -249,40 +173,54 @@ public void setParagraphDataset(Dataset<Row> rowDataset) {
...
-            currentAJAXLength = ajaxRequest.getLength();
-            currentSearchString = ajaxRequest.getSearch().getValue();
-            currentOrder = ajaxRequest.getOrder();
-        }
+        List<Order> currentOrder = null;
 
         // TODO these all decode the JSON, it refactor therefore to decode only once
         // searching
-        List<String> searchedList = DTSearch.search(datasetAsJSON, currentSearchString);
+        List<String> searchedList = DTSearch.search(datasetAsJSON, searchString);
```
eemhu: Other DT-prefixed objects should also be refactored to not include static methods

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNgTest.java
```diff
@@ -202,24 +205,76 @@ public void testAJAXResponse() {
...
     }
 
+    @Test
+    public void AjaxRequestToJsonTest(){
+        Gson gson = new Gson();
+
+        String paragraphId = "testParag";
+        String noteId ="testNoteId";
+        String angularObjectName = "AJAXRequest_"+paragraphId;
+        String angularObjectContent = "{\"start\":0,\"length\":25,\"search\":{\"value\":\"\",\"regex\":\"false\"}}";
```
eemhu: should this also be generated with jakarta?

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
```diff
@@ -1096,18 +1099,49 @@ public void onSuccess(Note newNote, ServiceContext context) throws IOException {
...
+  // -> updateAngularObject() updates an AJAXRequest angular object associated with a specific paragraph
+  // -> AJAXRequestWatcher catches the updated angular object
+  // -> values are passed to DTTableDatasetNG that does searching and pagination
+  // -> searched and paginated data is written via InterpreterContext
+  // -> Generates a PARAGRAPH_UPDATE_OUTPUT websocket event to be sent to the UI
+
+  private void updateParagraphResult(NotebookSocket conn,
+                                     ServiceContext context,
+                                     Message fromMessage) throws IOException {
+    final String noteId = (String) fromMessage.get("noteId");
```
eemhu: Can `get` return null?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: Created a new issue #234 for refactoring

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: Created issue https://github.com/teragrep/zep_01/issues/235 for refactoring

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: Looks like the reasons this catch exists at all is due to:
1. InterpreterOutput.write() can throw an IOException when it has an InterpreterOutputChangeListener object assigned to it (But currently that object is never assigned)
2. InterpreterOutput.flush() inherits the IOException from ByteArrayOutputStream.flush(). (But ByteArrayOutputStream.flush() can never throw an IOException since it's flush() method's body is empty)

Looks to me we could simply remove the IOExceptions from InterpreterOutput.write() and InterpreterOutput.flush(), so we don't have to catch them here.

It's out of scope for this PR, but I made a new issue for this refactoring: https://github.com/teragrep/zep_01/issues/236

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: Fixed in commit [97de6b1](https://github.com/teragrep/zep_01/pull/230/commits/97de6b1185ebf2d5e53f8de7f3f86ea83847f625)

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNgTest.java
Laukkala: Fixed in commit [0c30ab6](https://github.com/teragrep/zep_01/pull/230/commits/0c30ab6c394e0e5f98e18a93776981fbb7dad2db)

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
Laukkala: It can, in case frontend sends a malformed request that doesn't contain a "noteId" field. Same goes for "paragraphId".

Zeppelin is supposed to be able to handle nulls for noteIds and paragraphIds, but I added a null check with a specific exception message for these strings so that UI gets some feedback on their malformed requests in commit [b197d8f](https://github.com/teragrep/zep_01/pull/230/commits/b197d8f0c752380c0f1d187df798b829fab5bbe6)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
eemhu: should the `rowLists` be ctor arguments instead?

Laukkala: Added the rowLists to the constructor in this commit: [a100cab](https://github.com/teragrep/zep_01/pull/230/commits/a100cabf296d4b66d165635a3d2160e462b3366f)

I had to create new instances of the DT-prefixed objects during the call to searchAndPaginate(), because they are operating on a mutable variable "datasetAsJSON" (which is set in the method setParagraphDataset().
Refactoring the setParagraphDataset would be a pretty large undertaking, so it's probably not in scope for this issue.

Laukkala:
Added a ValidatedMessage decorator in commit [1ab1adc](https://github.com/teragrep/zep_01/pull/230/commits/1ab1adc5a9cd07add6608e60271af8a5e6e8a792)

The decorator ended up being quite complicated, since it needs to be able to validate the positions of the keys within the hierarchy.

In practice, the validation is comparing two JsonObjects with arbitrary Json contents, and returns true if one object contains every key that is present in the other object, in the same position.

If there is some premade method in the library such along the lines of JsonObject.containsAllKeys(JsonObject other), that would have been great

zeppelin-common/src/main/java/com/teragrep/zep_01/common/ValidatedMessage.java
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * @param toValidate
+   * @param toCompare
+   * @throws JsonException
+   */
+  private void validateObject(JsonObject toValidate, JsonObject toCompare) throws JsonException{
+    for (String requiredKey : toCompare.keySet()) {
+      JsonValue valueToValidate = toValidate.get(requiredKey);
+      JsonValue valueToCompare = toCompare.get(requiredKey);
+      if(valueToValidate == null){
```
eemhu: you can use containsKey instead of null check

eemhu:
The validatedMessage object does seem a bit complicated, was the plan to use it elsewhere as well?

Laukkala:
> The validatedMessage object does seem a bit complicated, was the plan to use it elsewhere as well?

Refactored ValidatedMessage with @eemhu, as it was only planned to use it in this one instance, so it need not be generic.

eemhu (commented):
Add `equals()` and `hashCode()` implementations and tests for the new objects, otherwise ok!

Laukkala:
> Add `equals()` and `hashCode()` implementations and tests for the new objects, otherwise ok!

Added equals and hashCode to new objects along with EqualsVerifier in commit [7d27a8e](https://github.com/teragrep/zep_01/pull/230/commits/7d27a8ef03af2f6b83c25410b6ad095000071575)

eemhu (dismissed):
LGTM

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTHeader.java
```diff
@@ -100,4 +108,29 @@ public static String schemaToHeader(StructType schema) throws TransformerExcepti
...
     }
+
+    public String json() {
+
+        JsonObjectBuilder builder = Json.createObjectBuilder();
+
+        Iterator<StructField> it = schema.iterator();
+        while(it.hasNext()) {
+            StructField column = it.next();
+            builder.add(column.name(),"");
```
kortemik: builder uses empty values for keys. is there a reason not to use JsonArrayBuilder?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -45,6 +45,7 @@
  */
 package com.teragrep.pth_07.ui.elements.table_dynamic;
 
+import com.google.gson.Gson;
```
kortemik: why Gson here and jakarta-json aka javax.json in DTHeader? I would use onky javax.json here

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -174,8 +117,12 @@ void refreshPage() {
     void handeAJAXRequest(AJAXRequest ajaxRequest) {
```
kortemik: nulls are checked already in AJAXRequestWatcher.

please refactor AJAXRequestWatcher to use jakarta-json and to produce null free object or stub if needed

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -174,8 +117,12 @@ void refreshPage() {
...
-
-            updatePage(ajaxRequest);
+            // Apply defaults if parameters are not provided
+            int start = (ajaxRequest.getStart() == null) ? 0 : ajaxRequest.getStart();
+            int length = (ajaxRequest.getLength() == null) ? 25 : ajaxRequest.getLength();
+            int draw = (ajaxRequest.getDraw() == null) ? 1 : ajaxRequest.getDraw();
+            String searchString = (ajaxRequest.getSearch().getValue() == null) ? "" : ajaxRequest.getSearch().getValue();
+            updatePage(start,length,searchString, draw);
         }
         catch (JsonSyntaxException e) {
```
kortemik: JsonSyntaxException is thrown in AJAXRequestWatcher not here why catching it here?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -96,74 +89,24 @@ public final class DTTableDatasetNg extends AbstractUserInterfaceElement {
...
-                getInterpreterContext().getAngularObjectRegistry().add(
-                "DTServerUpdate",
-                0,
-                getInterpreterContext().getNoteId(),
-                getInterpreterContext().getParagraphId(),
-                true
-        );
-
-        DTServerUpdateAngularObject.addWatcher(new DTServerUpdateWatcher(interpreterContext));
+        this.gson = new Gson();
```
kortemik: gson is not used here, why new Gson()?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -304,12 +243,15 @@ static JsonArray dataStreamParser(List<String> data){
         }
     }
 
-    static JsonObject DTNetResponse(JsonArray data, int ID, int length){
+    // Added headers to this object, should be contained in the AJAXResponse that UI receives.
+    static JsonObject DTNetResponse(JsonArray data, JsonObject datasetAsJSONSchema, int draw, int recordsTotal, int recordsFiltered){
         try{
             JsonObjectBuilder builder = Json.createObjectBuilder();
+            builder.add("headers",datasetAsJSONSchema);
```
kortemik: these are schemaHeadersAsXML. please at least rename the variable as such. as for the further analysis of the code: why data exchange is done in xml? it would be much of a better separation of concern to leave such tricks to UI as it requires the data format change.

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
```diff
@@ -249,40 +172,56 @@ public void setParagraphDataset(Dataset<Row> rowDataset) {
...
+        List<String> paginatedList = dtPagination.paginate(length, start);
 
         // ui formatting
         JsonArray formated = dataStreamParser(paginatedList);
-        JsonObject response = DTNetResponse(formated, currentAJAXID, orderedlist.size());
-        AJAXResponseAngularObject.set(response.toString(), true);
+
+        JsonObject headers;
+        // header formatting
+        if(datasetAsJSONFormattedSchema != ""){
```
kortemik: why first convert from dataset.schema which is type of structype to json using iterator and builder, then to string then back to json using a reader?

Dataset has method schema() which returns StructType which inherits from DataType which has json() method that produces json with on hit wonder call 

https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/Dataset.html#schema()

https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/types/DataType.html#json()

Wouldn't this do?

Perhaps DTHeader should be gone by then.

zeppelin-common/src/main/java/com/teragrep/zep_01/common/ValidatedMessage.java
```diff
@@ -0,0 +1,87 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.teragrep.zep_01.common;
+
+import com.google.gson.internal.LinkedTreeMap;
```
kortemik, line 20: why gson and jakarta-json are both present?

zeppelin-common/src/main/java/com/teragrep/zep_01/common/ValidatedMessage.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Objects;
+
+public final class ValidatedMessage {
+  private final Message messageToValidate;
+
+  /**
+   * Checks that the message provided contains all of the required keys for a PARAGRAPH_UPDATE_RESULT message. Returns true if message is valid, false if not.
+   * @param messageToValidate Message object to be validated
+   */
+  public ValidatedMessage(Message messageToValidate){
```
kortemik, line 31: this validates already the message which goes through the AjaxRequestWatcher, why double verification is done?

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
```diff
@@ -1096,18 +1101,58 @@ public void onSuccess(Note newNote, ServiceContext context) throws IOException {
...
+    if(validatedMessage.isValid()){
+      // Casting is required to get Message parameters in correct format, as GSON parses all numbers as Doubles, and Message.get() returns a generic Object.
+
+      final String noteId = (String) fromMessage.get("noteId");
+      final String paragraphId = (String) fromMessage.get("paragraphId");
+
+      // Build an interpreterGroupId based on given user and note Id.
+      // InterpreterGroupId is used to find the correct AngularObjectRegistry instance containing the DTTableDatasetNG object we want to pass the search, length, start and draw values to.
+
+      final String interpreterGroupId = "spark-"+conn.getUser()+"-"+noteId;
```
kortemik: i wonder if there is available such information ready made, building such is errorprone and does not work for others than spark

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
```diff
@@ -1096,18 +1101,58 @@ public void onSuccess(Note newNote, ServiceContext context) throws IOException {
...
+  // -> values are passed to DTTableDatasetNG that does searching and pagination
+  // -> searched and paginated data is written via InterpreterContext
+  // -> Generates a PARAGRAPH_UPDATE_OUTPUT websocket event to be sent to the UI
+
+  private void updateParagraphResult(NotebookSocket conn,
+                                     ServiceContext context,
+                                     Message fromMessage) throws IOException {
+    ValidatedMessage validatedMessage = new ValidatedMessage(fromMessage);
+    if(validatedMessage.isValid()){
+      // Casting is required to get Message parameters in correct format, as GSON parses all numbers as Doubles, and Message.get() returns a generic Object.
```
kortemik, line 1116: issue would be in place for removing such gson awkwardness

kortemik (changes requested):
quite much, @eemhu please look into the PRs more in detail

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: Schema.json() includes additional fields for each header, while we only need its name. Looks like the iterator was created to remove these fields first. Example of Schema.json() output for a schema with a single header:
`{"type":"struct","fields":[{"name":"_time","type":"timestamp","nullable":false,"metadata":{}}]}`

We can turn the JSONschema field from a String into a JsonArray to avoid double reading though. [a0a5dc7](https://github.com/teragrep/zep_01/pull/230/commits/a0a5dc7c32cce22173d059488a739096c5c45674)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTHeader.java
Laukkala: Changed this to JsonArray, as we only need keys.
[e0c9d80](https://github.com/teragrep/zep_01/pull/230/commits/e0c9d80eb76e707df8fc6024c856001093c1356a)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: Removed references to Gson from DTTableDatasetNg as it had no usages. Also removed JsonSyntaxException as it looks like a leftover from legacy code in commit [67f57e0](https://github.com/teragrep/zep_01/pull/230/commits/67f57e04bccd6b742ac06f11de3b0941083d3738)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: Removed JsonSyntaxException as it looks like a leftover from legacy code in commit [67f57e0](https://github.com/teragrep/zep_01/pull/230/commits/67f57e04bccd6b742ac06f11de3b0941083d3738)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: AjaxRequestWatcher now uses jakarta-json in commit [04ef0f8](https://github.com/teragrep/zep_01/pull/230/commits/04ef0f88427f215f902c4b6c67003a478589bf2b)

Replaced AjaxRequest object with a JsonObject instead, as it looks like the AjaxRequest class is simply a data holder that needed to exist to allow for Gson to deserialize JSON for validation purposes.

Along with other changes in this PR, AjaxRequest no longer has any usages outside tests, so it could be removed.

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: Leftover from a previous implementation, removed in commit [67f57e0](https://github.com/teragrep/zep_01/pull/230/commits/67f57e04bccd6b742ac06f11de3b0941083d3738) as it has no usages.

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTTableDatasetNg.java
Laukkala: Renamed variables to be more specific in commit [5f358af](https://github.com/teragrep/zep_01/pull/230/commits/5f358af6e947dff6c3afcf654996a1847f941313)

XML was used previously in draw() method to write headers directly to a paragraph, and having it be populated later by another update message. We removed draw() method to have the UI only receive data via one type of message, and headers are now being sent in JSON format

I don't think the XML formatted headers are used for anything anymore, so they were deleted in commit [2f56c1b](https://github.com/teragrep/zep_01/pull/230/commits/2f56c1bf36c49a97174b96122ec7eadded2a70c1)

zeppelin-common/src/main/java/com/teragrep/zep_01/common/ValidatedMessage.java
Laukkala: Leftover Exception declared in method signature. Removed as JsonExceptions are not thrown in current implementation. [ad8c309](https://github.com/teragrep/zep_01/pull/230/commits/ad8c30954cc399e94cb5a645e58d9adc73593719)

zeppelin-common/src/main/java/com/teragrep/zep_01/common/ValidatedMessage.java
Laukkala: Validation at the time the message is parsed is done so that we can respond with an error message if UI provides faulty parameters. 

I think AjaxRequestWatcher's validation would simply ignore invalid requests without any notification about faulty parameters, which might be confusing from the UI's perspective.

NotebookServer.onMessage() has a try-catch clause which is supposed to send an ERROR_INFO message to UI if an exception is encountered, but AjaxRequestWatcher's Exceptions are caught, logged and then discarded here: https://github.com/teragrep/zep_01/blob/1a0c747b26083de11d0064ff0f54a413740a1fbb/zeppelin-interpreter/src/main/java/com/teragrep/zep_01/display/AngularObject.java#L195 so Exceptions there cannot be sent to UI directly, at least without significant refactoring.

If we remove ValidatedMessage validation, UI will not be informed about bad requests, If we remove AjaxRequestWatchers validation, parsing a JsonObject can result in a NullPointerException, especially if we ever need to update AjaxRequest AngularObjects from another place apart from NotebookServer's Message objects. 

Should one or another of the validation steps be removed, and which one if so?

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
Laukkala: Will require replacing Message class's usage of GSON with jakarta-json, as well as refactoring serializeMessage() and deserializeMessage() methods. Created new issue #239

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/AJAXRequestWatcher.java
```diff
@@ -45,48 +45,50 @@
...
 
     public AJAXRequestWatcher(InterpreterContext context, DTTableDatasetNg dtTableDatasetNg) {
         super(context);
-        this.gson = new Gson();
         this.dtTableDatasetNg = dtTableDatasetNg;
     }
 
     @Override
-    public void watch(Object o, Object o1, InterpreterContext interpreterContext) {
+    public void watch(Object o, Object o1, InterpreterContext interpreterContext) throws JsonException {
```
kortemik: I think such broken request should return 400 (bad request) to the UI it is too late here to throw exception on the watch. i wonder if watch can do other than try to understand the request and if it does not, it can log a technical error about that such malformed was encountered and not proceed any further. it is the responsibility of the notifier not the one being notified that the notifications are correct. this is a common observer pattern here. could we change it so that if such a malformed request ends up here it would just be logged as i think throwing on the watch won't return 400 for the UI?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTHeader.java
```diff
@@ -61,10 +64,16 @@
...
-public class DTHeader {
+public final class DTHeader {
 
-    public static String schemaToHeader(StructType schema) throws TransformerException, ParserConfigurationException {
+    private final StructType schema;
+    public DTHeader(StructType schema){
+        this.schema = schema;
+    }
+
+    public String xml() throws TransformerException, ParserConfigurationException {
```
kortemik: why xml() is still present even though it is no longer used?

zeppelin-common/src/main/java/com/teragrep/zep_01/common/ValidatedMessage.java
kortemik: the way that the observer pattern is made here does not propagate the errors back like seen on the found snippet so it either requires refactoring of the observer pattern so that all exceptions get back here and error codes go back to the UI.

I would recommend that but to speed up the development a bit create an issue about that and validate it in the ValidatedMessage with the same logic that is in the watcher so we get the errors to the UI.

kortemik (changes requested):

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
Laukkala: InterpreterGroupId can be acquired from a given Paragraph: [e9a56f0](https://github.com/teragrep/zep_01/pull/230/commits/e9a56f07ca8f4038671f38c747f8a582a294b8da)
There are a lot of opportunities for zeppelin to return nulls while fetching the InterpreterGroupId. If an error occurs, a Bad Request response is sent to UI.

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/AJAXRequestWatcher.java
Laukkala: Added logging in commit [cb763c7](https://github.com/teragrep/zep_01/pull/230/commits/cb763c768bd7eea09ca5165263075e300b84abb6)
JsonException can occur when reading the request String into a JSON object using jakarta-json's Jsonreader. This should mean that the provided String is unparseable, so I added a catch clause for it too.

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/table_dynamic/DTHeader.java
Laukkala: Removed in commit [202bb1f](https://github.com/teragrep/zep_01/pull/230/commits/202bb1f019e7c397fb6f8079863cdeb3cd0b6182)

zeppelin-common/src/main/java/com/teragrep/zep_01/common/ValidatedMessage.java
Laukkala: Added issue #240 based on comments here.

In case of invalid data encountered in ValidatedMessage.isValid(), NotebookServer.updateParagraphResult() throws a BadRequestException, which should be caught by NotebookServer.onMessage() and result in an ERROR_INFO message being sent to UI.

kortemik (approved):

PR #244, opened 2025-09-03
Added a null check on RemoteAngularObject.set() when retrieving a process, as the process may be null 

### LINKED ISSUE #242: NullPointerException while changing pagination settings of a paragraph run by a different user.
**Describe the bug**
Changing pagination on a pargraph's datatable hangs, if the user is not the latest executor of the paragraph. Pagination only works when the same user who has executed the paragraph does the edits.

**Expected behavior**
Pagination should work even if someone else has run the paragraph.

**How to reproduce**
Have user1 run a paragraph, resulting in a Datatable. Have user2 change the pagination of the result.

**Software version**
ZEP_01 v 2.0.2.

Laukkala:
## Description
Fixed a NullPointerException that happens while changing DatatTable pagination settings on a paragraph that is owned by another user.

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality


zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/remote/RemoteAngularObject.java
```diff
@@ -46,9 +46,11 @@ public void set(Object o, boolean emitWeb, boolean emitRemoteProcess) {
 
     if (emitRemoteProcess) {
       // send updated value to remote interpreter
-      interpreterGroup.getRemoteInterpreterProcess().
-          updateRemoteAngularObject(
-              getName(), getNoteId(), getParagraphId(), o);
+      RemoteInterpreterProcess process = interpreterGroup.getRemoteInterpreterProcess();
```
MoonBow-1:
```suggestion
      final RemoteInterpreterProcess process = interpreterGroup.getRemoteInterpreterProcess();
```
`process` can be made final

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/remote/RemoteAngularObject.java
```diff
@@ -46,9 +46,11 @@ public void set(Object o, boolean emitWeb, boolean emitRemoteProcess) {
...
+      if(process != null){
+        process.updateRemoteAngularObject(
+                getName(), getNoteId(), getParagraphId(), o);
+      }
```
MoonBow-1, line 53: The solution makes sense to check if the process exists before trying to call its methods, but its a symptom of a larger issue, where the process **can** be null.

This technically doesn't check the `I have checked that my code does not have any NULL values.` code quality metric.

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/remote/RemoteAngularObject.java
MoonBow-1, line 1: I cannot see that this is tested anywhere, since now there is a possibility of not throwing an NPE

MoonBow-1 (changes requested):
I'm not quite familiar with ZEP_01, so I'm commenting more on the code quality

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/remote/RemoteAngularObject.java
Laukkala: Added a test in commit [f7ace22](https://github.com/teragrep/zep_01/pull/244/commits/f7ace22970fe654d84285babc5856784949324d6).

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/remote/RemoteAngularObject.java
Laukkala: fixed in commit [f7ace22](https://github.com/teragrep/zep_01/pull/244/commits/f7ace22970fe654d84285babc5856784949324d6)

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/remote/RemoteAngularObject.java
Laukkala: Nulls are due to Zeppelin's use of Lazy loading pattern, where processes are kept as null until some method initializes them by using ManagedInterpreterGroup.getOrCreateInterpreterProcess().

Perhaps a solution could be to create a Stub object for RemoteInterpeterProcess and use that to represent a process that has not been created yet, but due to possible side effects, I created a new issue https://github.com/teragrep/zep_01/issues/249 for it instead of implementing a stub object in this PR

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/remote/RemoteAngularObject.java
MoonBow-1: Test is now present, but uses mocking.
Due to my lack of knowledge, I'd wait for the architects review on this particular method of testing.

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/remote/RemoteAngularObject.java
MoonBow-1: `process` is now final

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/remote/RemoteAngularObject.java
MoonBow-1: This makes sense, and since a new issue is created, I'd resolve this if an architect is happy with this

MoonBow-1 (commented):
Added comments to previous review comments, and mostly looks good, but the mock in testing requires approval from an architect

kortemik (approved):

kortemik (changes requested):
please verify if interpreter is launched afterwards that it receives the set angular variables

Laukkala:
> please verify if interpreter is launched afterwards that it receives the set angular variables
If I understand the code correclty, the angular variables should be propagated to an interpreter that is launched afterwards as well:

RemoteInterpreterProcess.updateRemoteAngularObject() updates an object via Thrift in RemoteInterpreterServer's AngularObjectRegistry, if it exists.

It looks like RemoteInterpreter.open() copies the AngularObjects from its source InterpreterGroups Registry to the RemoteInterpreterServers AngularObjectRegistry, when the interpreter is first opened. 

Since we are calling this from NotebookService.updateAngularObject, where we are updating the source InterpreterGroups Registry, the updates should transfer over to RemoteInterpreterServers registry when the RemoteServer is started.

Laukkala:
Closing this as pagination is no longer relying on AngularObjects since PR: https://github.com/teragrep/zep_01/pull/263

PR #245, opened 2025-09-03
Added a configuration variable to SparkInterpreters, which allows DTTableDatasetNg to create JSON data in the format expected by the UI

Laukkala:
## Description
Added "spark.sql.jsonGenerator.ignoreNullFields = false" configuration value to enable including null values in Spark objects' .toJson() method. This overrides the default value of 'true' if the configuration value is not set.

This is done to ensure that DTTableDatasetNg can provide data to UI in the correct format expected by the UI.

Users should not change this configuration value, as doing so will cause UI to throw errors when missing keys are printed, but it is regardless available in interpreter-setting.json if for some reason changing it is necessary.

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

### DIFF
--- a/spark/interpreter/src/main/resources/interpreter-setting.json
+++ b/spark/interpreter/src/main/resources/interpreter-setting.json
@@ -143,6 +143,13 @@
         "defaultValue": true,
         "description": "Whether show the spark deprecated message, spark 2.2 and before are deprecated. Zeppelin will display warning message by default",
         "type": "checkbox"
+      },
+      "spark.sql.jsonGenerator.ignoreNullFields": {
+        "envName": null,
+        "propertyName": "spark.sql.jsonGenerator.ignoreNullFields",
+        "defaultValue": false,
+        "description": " Whether to discard null values when generating JSON from a Spark Dataset. Warning: toggling this to 'true' may cause issues with displaying data in paragraphs!",
+        "type": "checkbox"
       }
     },
     "editor": {
--- a/spark/interpreter/src/main/resources/interpreter-setting.json
+++ b/spark/interpreter/src/main/resources/interpreter-setting.json
@@ -148,7 +148,7 @@
         "envName": null,
         "propertyName": "spark.sql.jsonGenerator.ignoreNullFields",
         "defaultValue": false,
-        "description": " Whether to discard null values when generating JSON from a Spark Dataset. Warning: toggling this to 'true' may cause issues with displaying data in paragraphs!",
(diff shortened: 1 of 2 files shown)


spark/interpreter/src/main/resources/interpreter-setting.json
```diff
@@ -143,6 +143,13 @@
         "defaultValue": true,
         "description": "Whether show the spark deprecated message, spark 2.2 and before are deprecated. Zeppelin will display warning message by default",
         "type": "checkbox"
+      },
+      "spark.sql.jsonGenerator.ignoreNullFields": {
+        "envName": null,
+        "propertyName": "spark.sql.jsonGenerator.ignoreNullFields",
+        "defaultValue": false,
+        "description": " Whether to discard null values when generating JSON from a Spark Dataset. Warning: toggling this to 'true' may cause issues with displaying data in paragraphs!",
```
MoonBow-1:
```suggestion
        "description": "Whether to discard null values when generating JSON from a Spark Dataset. Warning: toggling this to 'true' may cause issues with displaying data in paragraphs!",
```
Small typo in the description

MoonBow-1:
From my limited knowledge, the UI expects that all the keys are present, but the value can be null?

Is there a reason for the values to be null?

MoonBow-1 (changes requested):
Small typo

spark/interpreter/src/main/resources/interpreter-setting.json
Laukkala: Fixed typo in commit [934012e](https://github.com/teragrep/zep_01/pull/245/commits/934012eeba7a137032aa3c3dc8321b23dff38787)

Laukkala:
> From my limited knowledge, the UI expects that all the keys are present, but the value can be null?
> 
> Is there a reason for the values to be null?

Yes, that's correct. UI expects to have every key available for every row of data, even if there is no data for that key in a row.
The absence of data is expected by the UI to be represented by a JSON null value.

I don't think DPL enforces that every row of data must always contain some value for every key, so I think it is possible to receive null values. We thought about including a default value instead of a null, but that would interfere with the actual data the customer is trying to analyze.

For example if we had a dataset of water temperature measurements from some sensors, but one sensor was malfuntioning when it was queried, resulting in the absence of data, we don't want to modify the dataset so that the malfunctioning sensor would output 0 degrees in the middle of summer.

kortemik (approved):

MoonBow-1 (approved):
LGTM

PR #263, opened 2025-10-02
Issue 262 Refactoring server-interpreter communication

Laukkala:
## Description

Refactored the communication between the server and interpreter processes so that they no longer depend on AJAXRequest AngularObjects. Now pagination and searching functions can be requested from the Interpreter via Thrift, which should increase reliability.

Exceptions should be propagated to the requester from the interpreterProcess, so the server should always respond to each request.

Some checklist items must be left unchecked due to coupling with other parts of ZEP_01
Not all methods of every object that has been modified has a test already and due to coupling, creating tests would be very time-consuming and may require refactoring.
#### General

- [ ] I have tested at least all public methods.

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

- [ ] I have tested [Object Equality](https://docs.oracle.com/javase/6/docs/api/java/lang/Object.html#equals%28java.lang.Object%29).

### Code Quality

- [ ] I have checked that my code follows metrics set in Procedure: Class Metrics.
- [ ] I have checked that my code follows metrics set in Procedure: Method Metrics.
- [ ] I have checked that my code follows metrics set in Procedure: Object Quality.
- [ ] I have checked that my code does not have any NULL values.


spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/DPLInterpreter.java
```diff
@@ -292,4 +292,26 @@ public Scheduler getScheduler() {
...
+        if(notebookParagraphUserInterfaceManager == null){
+            throw new InterpreterException("DPLInterpreter's notebookParagraphUserInterfaceManager map is not instantiated!");
+        }
+        if(!notebookParagraphUserInterfaceManager.containsKey(noteId)){
+            throw new InterpreterException("DPLInterpreter does not have a UserInterfaceManager for note id "+noteId);
+        }
+        if(!notebookParagraphUserInterfaceManager.get(noteId).containsKey(paragraphId)){
+            throw new InterpreterException("DPLInterpreter does not have a UserInterfaceManager for paragraph id "+paragraphId+" within note id "+noteId);
+        }
+        if(notebookParagraphUserInterfaceManager.get(noteId).get(paragraphId).getDtTableDatasetNg() == null){
```
StrongestNumber9: I think there is a possible TOCTOU issue with getting and re-getting them in multithreaded context as they could be potentially changed between check and get. Maybe store results of gets in local variables?

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
```diff
@@ -1100,60 +1092,77 @@ public void onSuccess(Note newNote, ServiceContext context) throws IOException {
...
-    }
-    else {
+    if(!validatedMessage.isValid()) {
       throw new BadRequestException("Request must contain \"noteId\", \"paragraphId\", \"start\", \"length\", \"draw\" and \"search.value\" parameters!");
     }
+    // Casting is required to get Message parameters in correct format, as GSON parses all numbers as Doubles, and Message.get() returns a generic Object.
+    final String msgId = fromMessage.msgId;
+    final String noteId = (String) fromMessage.get("noteId");
+    final String paragraphId = (String) fromMessage.get("paragraphId");
+    final int start = (int) Double.parseDouble(fromMessage.get("start").toString());
```
StrongestNumber9, line 1107: Why parse as double and then cast to int instead of parsing directly to int?

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
```diff
@@ -1100,60 +1092,77 @@ public void onSuccess(Note newNote, ServiceContext context) throws IOException {
...
+    final int length = (int) Double.parseDouble(fromMessage.get("length").toString());
+    final String search = (String) ((Map) fromMessage.get("search")).get("value");
+    final int draw = (int) Double.parseDouble(fromMessage.get("draw").toString());
+
+    Note note = getNotebook().getNote(noteId);
+    if(note == null){
+      throw new BadRequestException("No such note!");
```
StrongestNumber9: I would add more identifying information to these exceptions, as `No such note: 'ABCXYZ'` and `No such paragraph: 'paragraph_abc_xyz'` or so

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
```diff
@@ -1100,60 +1092,77 @@ public void onSuccess(Note newNote, ServiceContext context) throws IOException {
...
+              .put("data",dataset)
+              .put("index",0)
+              .put("noteid",noteId)
+              .put("paragraphId",paragraphId)
+              .put("type",InterpreterResult.Type.JSONTABLE);
+      conn.send(serializeMessage(msg));
+    }
+    catch (InterpreterException exception){
+      // Log the Exception to technical logs, only send a generic error message to UI.
+      LOG.error("Failed to access data from Interpreter process for note: " +noteId + " paragraph: " +paragraphId+ " cause: " + exception);
```
StrongestNumber9: Should use proper formatting instead of concatenation

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/DPLInterpreter.java
Laukkala: Stored the results into local variables in commit [a74f49c](https://github.com/teragrep/zep_01/pull/263/commits/a74f49c97d9f6fb1f09afb5659a756560615b932)

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
Laukkala: It's a fun quirk of GSON where all Json numbers are deserialized to Doubles. 
Message.get() also returns a generic Object. So numeric values must be first cast to Doubles, and then turned into integers.

Doing it like this could also work, but we still need to cast it twice since Object doesn't have .intValue():
int start = ((Double) fromMessage.get("start")).intValue()

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
Laukkala: Added ids to BadRequestExceptions in commit [46595e4](https://github.com/teragrep/zep_01/pull/263/commits/46595e4850699994b30d305ffaf07fb4e09ae9dc)

zeppelin-server/src/main/java/com/teragrep/zep_01/socket/NotebookServer.java
Laukkala: Changed to use formatting in commit [ecf8be1](https://github.com/teragrep/zep_01/pull/263/commits/ecf8be10b5c00fdc11da30d0a3bf1918b4e3a9a8)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/DPLInterpreter.java
```diff
@@ -292,4 +294,29 @@ public Scheduler getScheduler() {
...
+            throw new InterpreterException("DPLInterpreter does not have a UserInterfaceManager for paragraph id "+paragraphId+" within note id "+noteId);
+        }
+        DTTableDatasetNg dtTableDatasetNg = userInterfaceManager.getDtTableDatasetNg();
+        if(dtTableDatasetNg == null){
+            throw new InterpreterException("UserInterfaceManager for paragraph id "+paragraphId+" does not have a DTTableDatasetNG object!");
+        }
+        if(dtTableDatasetNg.getDatasetAsJSON().isEmpty()){
+            throw new InterpreterException("Dataset of paragraph "+paragraphId+" within note "+noteId+" is empty!");
+        }
+        JsonObject json = notebookParagraphUserInterfaceManager.get(noteId).get(paragraphId).getDtTableDatasetNg().SearchAndPaginate(draw,start,length,searchString);
```
StrongestNumber9: This should use the `dtTableDatasetNg` variable

Laukkala: Fixed in commit [3dda188](https://github.com/teragrep/zep_01/pull/263/commits/3dda1888f74b4313600d5f4f9a4bcd9ce9e5c38b)

kortemik (approved):

PR #286, opened 2026-01-27
Issue 285 Scala version without ProcessBuilder

### LINKED ISSUE #169: Finding scala version via spark-submit adds couple of seconds to each execution
**Description**

Use a way to detect wanted scala version via other routes than executing external binaries everytime.

**Use case or motivation behind the feature request**

https://github.com/teragrep/zep_01/blob/e233bda0c6396a0a4dabd69b63f9968a84e66c8f/zeppelin-zengine/src/main/java/org/apache/zeppelin/interpreter/launcher/SparkInterpreterLauncher.java#L247-L249

```
# time ./spark-submit --version 2> /dev/null

real    0m1.672s
user    0m2.567s
sys     0m0.246s
```
 
**Related issues**

Found while investigating a fix for #165

**Additional context**

Laukkala:
## Description
This removes the creation of a new process just for getting the proper Scala version, instead inferring the version from filenames of spark jars that are in Spark home directory.

Running of the process takes a few seconds, so inferring the version from file names should be a bit faster.

fixes #169


zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -308,14 +269,18 @@ public boolean accept(File dir, String name) {
     } else {
       // spark 2.x if spark/lib doesn't exists
       File sparkJarsFolder = new File(sparkHome + "/jars");
-      boolean sparkRepl211Exists =
-              Stream.of(sparkJarsFolder.listFiles()).anyMatch(file -> file.getName().contains("spark-repl_2.11"));
-      if (sparkRepl211Exists) {
-        return "2.11";
-      } else {
-        return "2.10";
+      Optional<File> sparkReplFound = Arrays.stream(sparkJarsFolder.listFiles()).filter(file -> file.getName().contains("spark-repl")).findFirst();
```
elliVM: this fixes the original issue which is good. but it introduces some other possible issues.

`findFirst()` relies on a single arbitary match, while in practice there is often only one `spark-repl` jar there can in some cases be multiple versions or shaded artifacts. we could end up picking the wrong one here.

we check for only 2.11 and 2.12 and will break for any other versions, since the version is encoded in the filename would it be possible to extract it instead using a pattern?

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -308,14 +269,18 @@ public boolean accept(File dir, String name) {
...
+        File sparkRepl = sparkReplFound.get();
+        if (sparkRepl.getName().contains("spark-repl_2.11")) {
+          return "2.11";
+        }
+        if (sparkRepl.getName().contains("spark-repl_2.12")) {
+          return "2.12";
+        }
       }
     }
+    throw new Exception("Could not detect a valid Scala version!");
```
elliVM: The interpreter will fail to start if a scala version detection fails here. Is the correct version strictly required for correct function or could this be just a warning with a fallback?

elliVM (changes requested):

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
Laukkala: I replaced this with a regex pattern in commit [8b5d62a](https://github.com/teragrep/zep_01/pull/286/commits/8b5d62a60db63e3c074a13468fc834425caf8b44)

I think multiple versions (or not having any matching jar) is an error case in previous implementations, so I added Exceptions for those.

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
Laukkala: I think not having a supported version would result in the Interpreter not operating properly, so failure to open seems like the correct aciton. I think the existence of these files are checked because each Spark version supports one or two different versions of Scala, and they should match before we try to run anything.

elliVM (approved):
LGTM

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -308,12 +273,26 @@ public boolean accept(File dir, String name) {
...
-              Stream.of(sparkJarsFolder.listFiles()).anyMatch(file -> file.getName().contains("spark-repl_2.11"));
-      if (sparkRepl211Exists) {
-        return "2.11";
-      } else {
-        return "2.10";
+      File[] fileArray = sparkJarsFolder.listFiles(file -> file.getName().contains("spark-repl"));
+      if (fileArray == null) {
+        throw new Exception("Failed to detect Scala version! An IO error occurred while accessing directory at: " + sparkJarsFolder);
+      }
+      List<File> replFiles = Arrays.asList(fileArray);
```
eemhu: any reason for the array to list conversion?

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -308,12 +273,26 @@ public boolean accept(File dir, String name) {
...
       File sparkJarsFolder = new File(sparkHome + "/jars");
-      boolean sparkRepl211Exists =
-              Stream.of(sparkJarsFolder.listFiles()).anyMatch(file -> file.getName().contains("spark-repl_2.11"));
-      if (sparkRepl211Exists) {
-        return "2.11";
-      } else {
-        return "2.10";
+      File[] fileArray = sparkJarsFolder.listFiles(file -> file.getName().contains("spark-repl"));
+      if (fileArray == null) {
+        throw new Exception("Failed to detect Scala version! An IO error occurred while accessing directory at: " + sparkJarsFolder);
```
eemhu: Would be preferred to use more detailed exception type

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
eemhu: I would think Spark 1.6 detection is no longer necessary
Spark 3.x and later 4.x should be the only supported ones?

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
```diff
@@ -308,12 +273,26 @@ public boolean accept(File dir, String name) {
...
+
+      Pattern pattern = Pattern.compile("spark-repl_(?<majorVersion>\\d*)\\.(?<minorVersion>\\d*)-.*\\.jar");
+      Matcher matcher = pattern.matcher(fileName);
+      if(matcher.matches()){
+        String majorVersion = matcher.group("majorVersion");
+        String minorVersion = matcher.group("minorVersion");
+        return majorVersion + "." + minorVersion;
+      }
+      else {
+        throw new Exception("Failed to detect Scala version! "+fileName+" in "+sparkJarsFolder+" uses an unsupported version! (expected 2.11 or 2.12) ");
```
eemhu: Should the actual version be checked here?

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
Laukkala: Not really, I'm probably just allergic to arrays. Fixed in commit [93ef5c4](https://github.com/teragrep/zep_01/pull/286/commits/93ef5c4d91d22bb5ca9174248dec0b328fec36df)

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
Laukkala: Now throws IOException instead of Exception, since it's being caught in calling method already. Fixed in commit [8c86f25](https://github.com/teragrep/zep_01/pull/286/commits/8c86f259db5736890fc655d14bc6498a7664b18d)

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
Laukkala: Rewrote this to use SPARK_HOME directory to detect major Spark version (spk_02 == spark 2, spk_03 == spark 3, etc...) and then check any spark jar for the scala version, as Spark 2 and Spark 3 can support scala versions 2.11, 2.12 or 2.13. Spark 4 seems to only use scala 2.13. rewritten in commit [e457e77](https://github.com/teragrep/zep_01/pull/286/commits/e457e77710452647ea6a976c500361fc4d5aa4a7)

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
Laukkala: I think the actual installed scala version is checked later [here](https://github.com/teragrep/zep_01/blob/7b02c42be7b1907866db23a98368a4d14168aadb/zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java#L136), and if the version returned by this method doesn't resolve to an installation directory, an error will get thrown.

zeppelin-zengine/src/main/java/com/teragrep/zep_01/interpreter/launcher/SparkInterpreterLauncher.java
eemhu: not sure if checking the spk_xx name is necessary there as it runs the same function anyways?

Laukkala: True, apparently Spark versions 2-4 should all have the same folder structure, where the jars can be found in the /jars/ folder, so as long as there are jar files available, we can get the scala version from their filenames.
Simplified the method in commit [2e9e75a](https://github.com/teragrep/zep_01/pull/286/commits/2e9e75a8727f97f94c2c899dc6fc0f81f99b77bd)

eemhu (approved):
LGTM

PR #303, opened 2026-03-18
Issue 301 Listener DPLMetricsListener threw an exception (NPE)

Laukkala:
## Description
Added null check to DPLMetricsListener as SQLExecutionUIData.metricValues() may return a null.

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

### DIFF
--- a/spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/DPLMetricsListener.java
+++ b/spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/DPLMetricsListener.java
@@ -86,18 +86,17 @@ public final class DPLMetricsListener extends StreamingQueryListener {
                 while (executionDataIterator.hasNext()) {
                     final SQLExecutionUIData executionData = executionDataIterator.next();
                     final scala.collection.Map<Object, String> scalaMetricValues = executionData.metricValues();
-                    if(scalaMetricValues != null){
-                    final Map<Object, String> metricValues = JavaConverters.mapAsJavaMap(scalaMetricValues);
-                    Seq<SQLPlanMetric> metrics = executionData.metrics();
-                    if(metrics != null){
-                        for (final SQLPlanMetric metric : JavaConverters.asJavaIterable(metrics)) {
+                    final Seq<SQLPlanMetric> scalaMetrics = executionData.metrics();
+                    if(scalaMetricValues != null && scalaMetrics != null){
+                        final Map<Object, String> metricValues = JavaConverters.mapAsJavaMap(scalaMetricValues);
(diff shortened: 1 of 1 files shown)


spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/DPLMetricsListener.java
```diff
@@ -86,11 +86,13 @@ public void onQueryProgress(final QueryProgressEvent event) {
                 while (executionDataIterator.hasNext()) {
                     final SQLExecutionUIData executionData = executionDataIterator.next();
                     final Map<Object, String> metricValues = JavaConverters.mapAsJavaMap(executionData.metricValues());
-                    for (final SQLPlanMetric metric : JavaConverters.asJavaIterable(executionData.metrics())) {
-                        final long id = metric.accumulatorId();
-                        final Object value = metricValues.get(id);
-                        if (metric.metricType().startsWith("v2Custom_") && value != null) {
-                            currentMetrics.put(metric.name(), value.toString());
+                    if(metricValues != null){
```
eemhu: should we check executionData.metricValues() != null instead ?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/DPLMetricsListener.java
```diff
@@ -86,11 +86,13 @@ public void onQueryProgress(final QueryProgressEvent event) {
...
+                        for (final SQLPlanMetric metric : JavaConverters.asJavaIterable(executionData.metrics())) {
```
eemhu: is same necessary for metrics() ?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/DPLMetricsListener.java
Laukkala: That shouldn't be necessary since JavaConverters.mapAsJavaMap() returns null if given null as a parameter, but I separated the null check to reduce method chaining in commit [d5f4455](https://github.com/teragrep/zep_01/pull/303/commits/d5f445524e038a7a3fab21494811cd5237010f7f)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/DPLMetricsListener.java
Laukkala: I'm not quite sure if metrics() is guaranteed to return something other than null, but I added a null check for it as well in commit [3748723](https://github.com/teragrep/zep_01/pull/303/commits/3748723c91714653470e42e90ee0471ea750c047)

eemhu (approved):

Issue #298, opened 2026-03-02
Remove SAVE_NOTE_FORMS message


Laukkala:
**Description**
SAVE_NOTE_FORMS is an obsoleted message, and should be removed from the server.

**Related issues**

**Additional context**

sryhanen:
This message is required currently for feature: https://docs.teragrep.com/teragrep-ui/1.0.0/languages/dynamic%20forms/scala-pyspark.html 

Related bug report: https://github.com/teragrep/ajs_01/issues/483

PR #308, opened 2026-03-30
Removed SAVE_NOTE_FORMS and REMOVE_NOTE_FORMS messages and their usages.

Laukkala:
## Description

Removed SAVE_NOTE_FORMS and REMOVE_NOTE_FORMS messages

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

### DIFF
--- a/zeppelin-server/src/main/java/com/teragrep/zep_01/service/NotebookService.java
+++ b/zeppelin-server/src/main/java/com/teragrep/zep_01/service/NotebookService.java
@@ -788,47 +788,6 @@ public class NotebookService {
     return cronUpdated;
   }
 
-  public void saveNoteForms(String noteId,
-                            Map<String, Object> noteParams,
-                            ServiceContext context,
-                            ServiceCallback<Note> callback) throws IOException {
-    if (!checkPermission(noteId, Permission.WRITER, Message.OP.SAVE_NOTE_FORMS, context,
-        callback)) {
-      return;
-    }
-
-    Note note = notebook.getNote(noteId);
-    if (note == null) {
-      callback.onFailure(new NoteNotFoundException(noteId), context);
-      return;
-    }
-
-    note.setNoteParams(noteParams);
-    notebook.saveNote(note, context.getAutheInfo());
-    callback.onSuccess(note, context);
-  }
-
-  public void removeNoteForms(String noteId,
-                              String formName,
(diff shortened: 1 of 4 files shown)


zeppelin-server/src/main/java/com/teragrep/zep_01/service/NotebookService.java
```diff
@@ -841,7 +800,7 @@ public NotebookRepoWithVersionControl.Revision checkpointNote(
       return null;
     }
 
-    if (!checkPermission(noteId, Permission.WRITER, Message.OP.REMOVE_NOTE_FORMS, context,
+    if (!checkPermission(noteId, Permission.WRITER, Message.OP.CHECKPOINT_NOTE, context,
```
Tiihott: It seems like it's safe to delete the Message.OP.X from the input parameters of the checkPermission() method, as the value is not used for anything by the method.

Tiihott (changes requested):
The checkPermission()-method in NotebookService.java should be cleaned up from unused input parameters.
Otherwise tests pass and changes look fine.

zeppelin-server/src/main/java/com/teragrep/zep_01/service/NotebookService.java
Laukkala: Removed input parameter in commit [323f4c5](https://github.com/teragrep/zep_01/pull/308/commits/323f4c58960c4838f4f891f87c54a1598e3b7531)

Tiihott (approved):
Tests pass and changes look ok.

Laukkala:
This message type is still required for now, but might become obsoleted in the future. See comment  https://github.com/teragrep/zep_01/issues/298#issuecomment-4244021237

eemhu:
@Laukkala PR still valid or should it be closed/set to draft mode?

Laukkala:
> @Laukkala PR still valid or should it be closed/set to draft mode?

I think the scope of the issue has changed. We have temporarily disabled dynamic forms from UI side, and it will be refactored into a new implementation, which will require changes to zep_01. I think this issue should also contain the replacement implementation, but the new dynamic forms have not been specified yet, so I think this issue should be left as a draft for now and updated when we implement a replacement for dynamic forms

Issue #283, opened 2026-01-15
Refactor ui content exchange format 


sryhanen:
**Description** 
Interface for requesting data in suitable formats for graphs is required.

When requesting pagination for server-side **raw data,** following message is sent from UI:
```
{
 op:"PARAGRAPH_UPDATE_RESULT",
 data: {
 noteId: <noteId>,
 paragraphId: <paragraphId>,
 draw: 2,
 start: 0,
 length:  1
}
```

Interface for the previous message should be extended so that with additional parameters:
```
{
 ...
 library: <type of library>, <-- type is alias for the plugin used to visualize data
 format: <type of format>, <-- graph type (linechart/barchart, etc.). 
}
```
**aggregated data** can be requested in different formats. In this case, start- and length-parameters are not required since we are not paginating the data. 

Additionally, aggregated data is currently sent as a string variable. 
-> This should be changed so that data is sent by default in array format.
Example of the default format: 
```
{
 headers:  string[] 
 data: object[][]
}
```

**Actions:**
- Implement support for requesting data in a plugin-specific format.
- Rename data types.

**Use case or motivation behind the feature request**
Required for [AJS_01 issue](https://app.zenhub.com/workspaces/teragrep-development-674473ee8a105f003111f8e9/issues/gh/teragrep/ajs_01/461)

**Additional context**
Currently data types are separated in following types: ELEMENT, TABLE, HTML, ANGULAR, TEXT, NETWORK, JSONTABLE,
where TABLE is alias for aggregated data and JSONTABLE for raw data.
Table should be renamed with better alias, json table should be named as raw data.

closes #234

kortemik:
I would change these so that from old `data` exhange we change to `type`, `options` and `content` where `type` determines the relevant `options` and `content`.

for raw results:

request
```json
{
  "op": "PARAGRAPH_UPDATE_RESULT",
  "data": {
    "noteId": "NEOOYECARF",
    "paragraphId": "paragraph_1766056017402_843761787",
    "draw": 2,
    "start": 0,
    "length": 25,
    "search": {
      "value": "",
      "regex": false,
      "fixed": []
    }
  },
  "principal": "userNameish",
  "ticket": "94d6b24b-761c-4f87-a3d6-e5eb097c811a",
  "roles": "[\"some\"]",
  "msgId": "klnbh-35"
}

```

response
```json
{
  "op": "PARAGRAPH_UPDATE_OUTPUT",
  "data": {
[...]

PR #319, opened 2026-04-16
Issue 284 performance updates as spark datasets

Laukkala:
## Description

This PR introduces support for storing performance data of a DPL query into a Spark dataset.

While a query is running, all performance updates received via DPLMetricsListener are now stored into a Dataset, and its contents are provided to UI in JSON basic format. Previously, only the latest performance update was sent to UI as a String message.

Performance data can be formatted using the same methods as #283 is using to format output. Pending the merging of issue #283, the formatting used is simply a JSON representation of all the performance data.
 
Since the performance data is in a Spark Dataset, we can use any of Spark's transformations to further refine the performance data based on a request from the UI if needed. This functionality has not yet been specified, so it falls outside of this PR for now.

## Checklists

Includes todos regarding formatting, which require issue #283 to be merged

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

- [ ] I have checked my code does not contain FIXME or TODO comments.


spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/Timestamp.java
eemhu, line 1: Could all of these /metric/* classes implement MetricValue<T> ? Add the common methods to the interface

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLMetricsListener.java
```diff
@@ -79,29 +84,35 @@ public void onQueryStarted(final QueryStartedEvent event) {
...
                     final Map<Object, String> metricValues = JavaConverters.mapAsJavaMap(executionData.metricValues());
-                    for (final SQLPlanMetric metric : JavaConverters.asJavaIterable(executionData.metrics())) {
-                        final long id = metric.accumulatorId();
-                        final Object value = metricValues.get(id);
-                        if (metric.metricType().startsWith("v2Custom_") && value != null) {
-                            currentMetrics.put(metric.name(), value.toString());
+                        for (final SQLPlanMetric metric : JavaConverters.asJavaIterable(executionData.metrics())) {
+                            final long id = metric.accumulatorId();
+                            final String value = metricValues.get(id);
+                            if (metric.metricType().startsWith("v2Custom_") && value != null && value != "null") {
```
eemhu: `!"null".equals(value);`

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLPerformanceEntry.java
```diff
@@ -0,0 +1,637 @@
...
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+public final class DPLPerformanceEntry {
+    private final RowsReadFromArchive rowsReadFromArchive;
```
eemhu: could these be a map instead?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/PerformanceIndicator.java
```diff
@@ -92,22 +104,19 @@ public void emit() {
...
-            newMessage.append("Metrics:\n");
-            for (final Map.Entry<String, String> metric : currentMetrics.entrySet()) {
-                newMessage.append(metric.getKey()).append(": ").append(metric.getValue()).append("\n");
-            }
+    public void sendPerformanceUpdate(){
+        // TODO: format dataset to uPlot format (requires ZEP_01#283)
+        // TODO: format into DataTables format as well if needed (requires ZEP_01#283)
+        final JsonArrayBuilder output = Json.createArrayBuilder();
+        final List<String> rows = performanceData.get().toJSON().collectAsList();
+        for (String row : rows) {
```
eemhu: I think you can avoid parsing JSON by using the following

```java
final Row r = ds.toJSON().agg(functions.collect_list(functions.col("value")).as("values")).first();
```

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLPerformanceEntry.java
Laukkala: Refactored in commit [d12c223](https://github.com/teragrep/zep_01/pull/319/commits/d12c2235ccd493c1aac7cb459b203f1c3c966a46)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/ui/elements/PerformanceIndicator.java
Laukkala: Removed collectAsList() call in commit [05b0c6e](https://github.com/teragrep/zep_01/pull/319/commits/05b0c6ed9a6443bc7e297a092a20f9d2ee3d897d). Had to do some more JSON parsing to avoid leaving escape characters to keys

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLMetricsListener.java
Laukkala: fixed in commit [f98cf9c](https://github.com/teragrep/zep_01/pull/319/commits/f98cf9c51c01d1423725b1369c9013406758862b)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/Timestamp.java
Laukkala: Currently these Metric objects can exist even though they have no value, so I added the MetricValue interface so that I can assign a stub object instead of a null.

I thought about making the Metrics Stubable, and having each a metric with no value be a Stub, but I needed access to the metric names, descriptions and StructFields before I had a value to assign, so only the Value of the metric should be a Stub.

I ended up creating a PerformanceSchema object which contains a collection of PerformanceMetrics, each with a stub value. This is used by PerformanceEntry to define the order of the metrics, and is compared to when new data is being entered to one of the metrics. PerformanceEntry also keeps a map of valid metrics that have values assigned to them.

It feels like there's some overlap between PerformanceSchema and PerformanceEntry, but I can't think of a  simpler solution for now at least

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/value/MetricValue.java
```diff
@@ -0,0 +1,52 @@
...
+ * To the extent this program is licensed as part of the Commercial versions of
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth_07.performance.metric.value;
+
+import com.teragrep.stb_01.Stubable;
+
+public interface MetricValue<T> extends Stubable {
+    public abstract java.lang.Object value();
```
eemhu: Generic type `T` is not used in interface

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/value/StubMetricValue.java
```diff
@@ -0,0 +1,75 @@
...
+
+public final class StubMetricValue<T> implements MetricValue<T> {
+
+    @Override
+    public boolean isStub() {
+        return true;
+    }
+
+    @Override
+    public java.lang.Object value() {
```
eemhu: `Object` -> `T`

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/ArchiveCompressedBytesProcessed.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+    private ArchiveCompressedBytesProcessed(final MetricValue<Long> value){
+        this.value = value;
+    }
+    public MetricValue<Long> metricValue() {
+        return value;
+    }
+
+    @Override
+    public ArchiveCompressedBytesProcessed withValue(java.lang.Object value) throws IncompatibleValueException {
```
eemhu: should `PerformanceMetric` also be of generic type?

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/RowsReadFromArchive.java
```diff
@@ -0,0 +1,129 @@
...
+import com.teragrep.pth_07.performance.metric.value.StubMetricValue;
+import com.teragrep.zep_01.common.exception.IncompatibleValueException;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.Metadata;
+import org.apache.spark.sql.types.StructField;
+
+import java.util.Objects;
+
+public final class RowsReadFromArchive implements PerformanceMetric {
```
eemhu: I'm thinking here, could these all be PerformanceMetricImpls instead, with name, description, value type and value defined in ctor? this way all of them could be one generic object and reduce duplicate code.

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/performance/metric/PerformanceSchemaTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+class PerformanceSchemaTest {
+
+    @Test
+    void testDefaultMetrics() {
+        PerformanceSchema performanceSchema = new PerformanceSchema();
+        Assertions.assertEquals(17,performanceSchema.metrics().size());
+
+        // Default values should all be stubs
+        for (int i = 0; i < 17; i++) {
```
eemhu: assert loops executed

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/value/MetricValue.java
Laukkala: Fixed in commit [8b9107e](https://github.com/teragrep/zep_01/pull/319/commits/8b9107e0187ab9583fb3fa17e37659afab6e62e3)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/value/StubMetricValue.java
Laukkala: Fixed in commit [8b9107e](https://github.com/teragrep/zep_01/pull/319/commits/8b9107e0187ab9583fb3fa17e37659afab6e62e3)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/ArchiveCompressedBytesProcessed.java
Laukkala: Changed in commit [e7c2eb4](https://github.com/teragrep/zep_01/pull/319/commits/e7c2eb4210fd064a8e42ecb783059848f488353d)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/RowsReadFromArchive.java
Laukkala: Refactored to use only one PerformanceMetric<T> object instead in commit [bf64f9b](https://github.com/teragrep/zep_01/pull/319/commits/bf64f9b8f8187f83f445a4cd3325db583e2e1b65).

PerformanceSchema now takes a list of PerformanceMetric objects which is used to define how the Dataset is structured. No parameter constructor creates the default metrics, but you can pass your own list if you want to define a different schema.

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/performance/metric/PerformanceSchemaTest.java
Laukkala: fixed in commit [8565169](https://github.com/teragrep/zep_01/pull/319/commits/85651696885b6785bda90ed899146c82405e4306)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/ArchiveCompressedBytesProcessed.java
eemhu: could withValue(Object value) be changed to withValue(T value) to avoid instanceof checks ?

Laukkala: If we changed the method to withValue(T value), we wouldn't be able to use withValue using Strings like in this [test](https://github.com/teragrep/zep_01/pull/319/changes#diff-c7b75d7ba2ded10eb8903973740ca93bac2b3c77e5e6aa54aed4a6bf735445c9R67)
We need to be able to pass Strings to PerformanceMetrics because SQLExecutionUIData.metricValues() ([used here](https://github.com/teragrep/zep_01/pull/319/changes#diff-95d238cb9a29ee79c5b7952764cb9175335e289053dda97aa784a3432bc3a40fR100)) returns a map of metricId keys and String values, even though the values might actually represent numerical data. We need to parse these into actual numerical values so that the Dataset we store them in can make calculations on them later.

DPLMetricsListener might also pass longs, integers and doubles to PerformanceMetrics (BatchId, EPS, etc)

Is it possible to have a method return a PerformanceMetric<Long> from  withValue(T value) by passing a String as the value? I'm not that familiar with generics, but from my testing if I call PerformanceMetric<Long>.withValue(String value), i get an error where value is expected to be Long type.

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/value/StubMetricValue.java
```diff
@@ -45,17 +45,19 @@
  */
 package com.teragrep.pth_07.performance.metric.value;
 
+import org.apache.spark.sql.types.DataType;
+
 import java.util.Objects;
 
-public final class StubMetricValue<T> implements MetricValue<T> {
+public final class StubMetricValue implements MetricValue {
```
eemhu: change stub from prefix to suffix

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLPerformanceEntry.java
```diff
@@ -46,49 +46,125 @@
...
+        }
+        try{
+            PerformanceMetric metric = metrics.get(key);
+            PerformanceMetric modifiedMetric = metric.withValue(value);
+            Map<String, PerformanceMetric> modifiedMetrics = new HashMap<>(metrics);
+            modifiedMetrics.put(key,modifiedMetric);
+            return new DPLPerformanceEntry(modifiedMetrics);
+        }
+        catch (IncompatibleValueException incompatibleValueException){
+            LOGGER.warn("Failed to update Query performance data for metric [{}] due to mismatched data type ",key,incompatibleValueException);
```
eemhu: log should be `<[{}]>`

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLPerformanceEntry.java
```diff
@@ -46,49 +46,125 @@
...
-                break;
-            }
+    /**
+     * Creates a new instance of this entry, with an updated value for the metric identified by key. If given key does not match with any of the metrics, returns this entry without modification.
+     * @param key PerformanceMetric to update. Must match with the key of one of the PerformanceMetrics this object was initialized with.
+     * @param value new value
+     * @return a modified instance of this DPLPerformanceEntry, with the PerformanceMetric identified by key having it's value replaced by the given value.
+     */
+    public DPLPerformanceEntry withData(final String key, final long value) {
+        if(!metrics.containsKey(key)){
```
eemhu: try to use single return statement

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/performance/DPLPerformanceEntryTest.java
```diff
@@ -91,52 +99,83 @@ void testWithDataIgnoresUnknownKeys() {
...
+                .master("local[*]")
+                .getOrCreate();
+
+        List<Row> rows = new ArrayList<Row>();
+        rows.add(entry1.asRow(entry1.performanceSchema()));
+        rows.add(entry2.asRow(entry1.performanceSchema()));
+        rows.add(entry3.asRow(entry1.performanceSchema()));
+
+        Dataset<Row> dataset = sparkSession.createDataFrame(rows,entry1.performanceSchema());
+        System.out.println("dataset");
```
eemhu: remove printlns

eemhu: also test missing assertions

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLPerformanceEntry.java
```diff
@@ -46,49 +46,125 @@
...
 
+    /**
+     * DPLPerformanceEntry represents a row of performance data from a DPL query. Each batch received from DPL corresponds to one DPLPerformanceEntry.
+     * DPLPerformanceEntry keeps a map of PerformanceMetrics, each of which can be updated using the withData() methods.
+     * DPLPerformanceEntry can produce a Spark schema that contains all the information for each of the PerformanceMetrics it contains.
+     * DPLPerformanceEntry is capable of turning itself into a Spark Row that can be added to a Spark Dataset.
+     */
     public DPLPerformanceEntry(){
-        this(new PerformanceSchema(),new HashMap<>());
+        this(Stream.of(
```
eemhu: I would just make this one-liner ctor into a separate object.

For example

MetricsSchema interface
DefaultMetricsSchema implementation

interface has asMap() method which returns the map

asMap could be implemented as such:
```java
public Map<String,PerformanceMetric> asMap() {
  final Map<String, PerformanceMetric> metrics = new HashMap<>();
  metrics.put("ArchiveCompressedBytesProcessed", new PerformanceMetric(...));
  // all the metrics ...
  return metrics;
}
```

then the DPLPerformanceEntry would take a MetricsSchema implementing object as the parameter.

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/metric/value/StubMetricValue.java
Laukkala: Fixed in commit [18cf29e](https://github.com/teragrep/zep_01/pull/319/commits/18cf29e237797eb6f6a2f5921a415e442d9a8e4a)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLPerformanceEntry.java
Laukkala: fixed in commit [a8668e8](https://github.com/teragrep/zep_01/pull/319/commits/a8668e8aa40545e084e3cf3566fae12651706f65)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLPerformanceEntry.java
Laukkala: fixed in commit [69583bb](https://github.com/teragrep/zep_01/pull/319/commits/69583bba129919e25d7732e13649b8523591fe22)

spark/dpl-interpreter/src/test/java/com/teragrep/pth_07/performance/DPLPerformanceEntryTest.java
Laukkala: fixed in commit [c01cd9c](https://github.com/teragrep/zep_01/pull/319/commits/c01cd9c125ad3c430f7561b77768c8b158298fd6)

spark/dpl-interpreter/src/main/java/com/teragrep/pth_07/performance/DPLPerformanceEntry.java
Laukkala: Created a new object in commit [750186d](https://github.com/teragrep/zep_01/pull/319/commits/750186dac1472d9dadd65a0b5f9a6a9750f30b71)

eemhu (approved):

PR #324, opened 2026-05-14
regex interpreter

kortemik:
#### General

- [ ] I have checked that my test files and functions have meaningful names.
- [ ] I have checked that each test tests only a single behavior.
- [ ] I have done happy tests.
- [ ] I have tested only my own code.
- [ ] I have tested at least all public methods.

#### Assertions

- [ ] I have checked that my tests use assertions and not runtime overhead.
- [ ] I have checked that my tests end in assertions.
- [ ] I have checked that there is no comparison statements in assertions.
- [ ] I have checked that assertions are in tests and not in helper functions.
- [ ] I have checked that assertions for iterables are outside of for loops and both sides of the iteration blocks.
- [ ] I have checked that assertions are not tested inside consumers.

#### Testing Data

- [ ] I have tested algorithms and anything else with the possibility of unbound growth.
- [ ] I have checked that all testing data is local and fully replaceable or reproducible or both.
- [ ] I have checked that all test files are standalone.
- [ ] I have checked that all test-specific fake objects and classes are in the test directory.
- [ ] I have checked that my tests do not contain anything related to customers, infrastructure or users.
- [ ] I have checked that my tests do not contain non-generic information.
- [ ] I have checked that my tests do not do external requests and are not privately or publicly routable.

#### Statements

- [ ] I have checked that my tests do not use throws for exceptions.
- [ ] I have checked that my tests do not use try-catch statements.
- [ ] I have checked that my tests do not use if-else statements.

#### Java

- [ ] I have checked that my tests for Java uses JUnit library.
- [ ] I have checked that my tests for Java uses JUnit utilities for parameters.

#### Other

- [ ] I have only tested public behavior and not private implementation details.
- [ ] I have checked that my tests are not (partially) commented out.
- [ ] I have checked that hand-crafted variables in assertions are used accordingly.
- [ ] I have tested [Object Equality](https://docs.oracle.com/javase/6/docs/api/java/lang/Object.html#equals%28java.lang.Object%29).
- [ ] I have checked that I do not have any manual tests or I have a valid reason for them and I have explained it in the PR description.

### Code Quality

- [ ] I have checked that my code follows metrics set in Procedure: Class Metrics.
- [ ] I have checked that my code follows metrics set in Procedure: Method Metrics.
- [ ] I have checked that my code follows metrics set in Procedure: Object Quality.
- [ ] I have checked that my code does not have any NULL values.
- [ ] I have checked my code does not contain FIXME or TODO comments.


regex/src/test/java/com/teragrep/zep_01/regex/RegexInterpreterTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.Properties;
+
+import static org.junit.Assert.assertEquals;
+
+public class RegexInterpreterTest {
+
+  private static RegexInterpreter java;
```
Laukkala: This should probably be called 'regexInterpreter' instead of 'java', used as such in testStaticRepl()

regex/src/test/java/com/teragrep/zep_01/regex/RegexInterpreterTest.java
```diff
@@ -0,0 +1,104 @@
...
+    InterpreterResult res = regexInterpreter.interpret(writer.toString(), context);
+
+    assertEquals(InterpreterResult.Code.SUCCESS, res.code());
+    assertEquals(InterpreterResult.Type.TEXT, res.message().get(0).getType());
+  }
+
+
+  @Test
+  public void testEmpty() {
+    InterpreterResult res = java.interpret("", context);
```
Laukkala: regexInterpreter.interpret

regex/src/test/java/com/teragrep/zep_01/regex/RegexInterpreterTest.java
```diff
@@ -0,0 +1,104 @@
...
+    StringWriter writer = new StringWriter();
+    PrintWriter out = new PrintWriter(writer);
+    out.println("^(?<line>.*)$");
+    out.println("content that matches really well to this regex");
+    out.close();
+
+    InterpreterResult res = regexInterpreter.interpret(writer.toString(), context);
+
+    assertEquals(InterpreterResult.Code.SUCCESS, res.code());
+    assertEquals(InterpreterResult.Type.TEXT, res.message().get(0).getType());
```
Laukkala, line 108: Should assert that res.message().get(0).toString() matches with expected output

regex/src/test/java/com/teragrep/zep_01/regex/RegexInterpreterTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Test
+  public void testEmpty() {
+    InterpreterResult res = java.interpret("", context);
+
+    assertEquals(InterpreterResult.Code.ERROR, res.code());
+    assertEquals("expected regex on the first line", res.message().get(0));
+
+  }
+
```
Laukkala, line 121: Add test case for non-matching content (Type should be ERROR, output should contain "does not match pattern")

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
```diff
@@ -0,0 +1,176 @@
...
+  public InterpreterResult interpret(String prompt, InterpreterContext context) {
+
+    final InterpreterResult rv;
+
+    int newlineIndex= prompt.indexOf('\n');
+    if (newlineIndex != -1) {
+      String regex = prompt.substring(0, newlineIndex);
+      String content = prompt.substring(newlineIndex + 1);
+
+      Pattern pattern = Pattern.compile(regex);
```
Laukkala: This can throw PatternSyntaxException if given an invalid regex. eg. this one with unclosed parentheses: "(invalidPattern"

Should be caught and produce an error message informing user that pattern was malformed

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
```diff
@@ -0,0 +1,176 @@
...
+      Matcher matcher = pattern.matcher(content);
+
+      InterpreterResult rvLocal;
+      try {
+        // java 11 does not have namedGroups as public so reflection is needed
+        Method namedGroupsMethod = Pattern.class.getDeclaredMethod("namedGroups");
```
Laukkala: Accessing protected fields via reflection should be avoided, this requires some JVM flags to be set so that it doesn't throw an InaccessibleObjectException, and it's not obvious to users that such flags should be set, so this should be a last resort solution.

The only other option I can think of is to parse the given regex pattern itself for the group names, but generally validating user input is error prone, so maybe reflection is the only valid option.

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
```diff
@@ -0,0 +1,176 @@
...
+
+        namedGroupsMethod.setAccessible(true);
+
+        @SuppressWarnings("unchecked")
+        Map<String, Integer> groupMap = (Map<String, Integer>) namedGroupsMethod.invoke(pattern);
+
+
+        if (matcher.find()) {
```
Laukkala: This will only find the first match in the regex, there might be more. Should be changed to a while loop.

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
```diff
@@ -0,0 +1,176 @@
...
+          JsonObjectBuilder builder = Json.createObjectBuilder();
+
+          for (String key : groupMap.keySet()) {
```
Laukkala: Currently RegexInterpreter doesn't match anything if it's not a named group. Is this wanted behaviour or should it also handle more basic regex use cases?

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
```diff
@@ -0,0 +1,215 @@
...
+        final String value = matcher.group(key);
+
+        if (value == null) {
+          recordSchemaDatumBuilder.addNull(key);
+        }
+        else {
+          recordSchemaDatumBuilder.add(key, value);
+        }
+        recordSchemaDatumBuilder.addNull("columnDescription");
+        recordSchemeDataBuilder.add(recordSchemaDatumBuilder.build());
```
Laukkala: Column description is always null, should there be some extracted data that gets placed here?

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
```diff
@@ -0,0 +1,215 @@
...
+        @SuppressWarnings("unchecked")
+        final Map<String, Integer> groupMapLocal = (Map<String, Integer>) namedGroupsMethod.invoke(pattern);
+        groupMap = groupMapLocal;
+      }
+      catch (InvocationTargetException | IllegalAccessException e) {
+        LOGGER.error("reflection error invoke", e);
+        throw new RegexInterpreterException("reflection error invoke", e);
+      }
+
+      if (!matcher.matches()) {
```
Laukkala: matcher.matches() requires that the entire content matches with the given regex pattern, and multiple matches within a group are not supported.

Take for example this pattern that creates a group called 'word' that contains every word in the content separated by a '-' character: `(?<=-|^)(?<word>\w+)(?=-|$)`. 
For the content `words-separated-by-dashes`, matcher.matches() would return false.  In this case we'd have to use matcher.find("word") instead, and the first call to matcher.find() would only return `words`. We'd have to use a while loop to get access to every value.

Should this kind of use case be supported, or will the payloads this interpreter will use conform to some stricter structure?

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
Laukkala: If the result JSON object will contain more data than this and the format needs to be more strictly defined, it might be good to create a new object to represent the format we are building here and move this logic to that object

Laukkala (commented):
Needs a few more test cases for different regex patterns and payloads to cover all expected success and failure cases.

Perhaps a few objects could be created that define and validate the format we are expecting the given regex pattern to conform to, as well as an object that defines what kind of JSON output RegexInterpreter gives out. 

Currently they are all in the interpret() method, and if the logic expands, it might get hard to read.

Uses reflection to access group names given by user, but there doesn't seem to be easy workarounds for it at present.

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  @Override
+  public void close() {
+
+
+  }
+
+  @Override
+  public InterpreterResult interpret(String prompt, InterpreterContext context) {
```
Laukkala, line 79: interpret() method has many responsibilities.

1.  Validate and parse `prompt` into `regex` and `content` strings
2.  Create a matcher based on `regex` and parse the names of named groups
3.  Apply `content` to matcher and extract matches
4.  Create a JSON string containing the matches found in step 3
5.  Return an InterpreterResult that indicates either a successful or unsuccessful with an attached payload containing either the successful result or an error message 

Consider splitting them up into at least private methods, if not separate objects.

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
kortemik: on java 20+ it would be possible to use namedGroups() but java 20+ is not in zep_01 so there is not other option than this hack.

kortemik:
added some more objecst

regex/src/main/java/com/teragrep/zep_01/regex/captureGroup/TextImpl.java
```diff
@@ -0,0 +1,68 @@
...
+ * Licensee must indemnify licensors and authors for any liability that these
+ * contractual assumptions impose on licensors and authors.
+ *
+ * To the extent this program is licensed as part of the Commercial versions of
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.zep_01.regex.captureGroup;
+
+public class TextImpl implements Text
```
eemhu, line 48: equals and hashCode implementations seem to be missing from a few objects.

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
```diff
@@ -0,0 +1,137 @@
...
+    try {
+      SkipablePrompt skipablePrompt = new SkipablePrompt(prompt);
+      SpliatblePrompt spliatblePrompt = new SpliatblePrompt(skipablePrompt.skipFirstLine());
```
eemhu: typo, Splitable

regex/src/main/java/com/teragrep/zep_01/regex/RegexInterpreter.java
```diff
@@ -0,0 +1,137 @@
...
+  @Override
+  public void close() {
+
+
+  }
+
+  @Override
+  public InterpreterResult interpret(String prompt, InterpreterContext context) {
+    try {
+      SkipablePrompt skipablePrompt = new SkipablePrompt(prompt);
```
eemhu: Skippable?

Laukkala (commented):
Looks good, but could use a few more tests for the more complicated methods.
Could potentially change use of Jsonable in places to more specific GroupJsonable if there is no requirement to use the more general interface.

PR #328, opened 2026-05-22
Issue 126 allow multiple notebooks with same name

### LINKED ISSUE #126: User should be able to create/import notebooks with the same name
And get a separate comprehensive error message in case operation was not successful
Connected AJS_01 issue: 
https://github.com/teragrep/ajs_01/issues/119

Laukkala:
## Description
Changed NoteManager so that it uses NoteId to identify Notebooks instead of paths. The actual files where notebooks are saved will have IDs appended, so there shouldn't be any conflicts with pathnames even if the same name is used.

fixes #126 

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality


zeppelin-server/src/main/java/com/teragrep/zep_01/service/NotebookService.java
```diff
@@ -1055,7 +1055,7 @@ public void moveFolderToTrash(String folderPath,
     LOGGER.info("Move folder {} to trash", folderPath);
 
     String destFolderPath = "/" + NoteManager.TRASH_FOLDER + "/" + folderPath;
-    if (notebook.containsNote(destFolderPath)) {
+    if (notebook.containsFolder(destFolderPath)) {
```
Tiihott, line 1058: Was this a bug that was fixed? The method used was changed from containsNote() to containsFolder().

Tiihott (changes requested):
Tests pass and changes look overall ok. Just need some clarification on the changes done to moveFolderToTrash() method.

zeppelin-server/src/main/java/com/teragrep/zep_01/service/NotebookService.java
Laukkala: Yes, the previously used method would not have found any notebooks since the path would correspond to a folder. I think the trash folder feature is currently disabled on the frontend anyway. 

Turns out the legacy test class didn't even include a test case for this feature so I added it in commit  [4096bcf](https://github.com/teragrep/zep_01/pull/328/commits/4096bcf1bf4bb7652967569cc8c5dc0b35a799ef)

Tiihott (dismissed):
LGTM

zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/NoteManager.java
```diff
@@ -469,12 +437,29 @@ public Map<String, Folder> getFolders() {
...
+            return subfolder.getNote(noteId);
+          }
+          // Carry on until every subfolder is searched.
+          catch (NoteNotFoundException noteNotFoundException){
+            continue;
+          }
+        }
+      }
+      // If the given noteId was not found in this or any of the subfolders, throw an Exception.
+      throw new NoteNotFoundException("Could not find note with id " + noteId);
```
eemhu: Try to see if this could be a single return value method.

eemhu:
how is the id generated, is there a chance that it could collide with another?

Laukkala:
> how is the id generated, is there a chance that it could collide with another?

Id is generated in IdHashes.generateId(), which takes a long as an input and uses encode() in the same class to generate an id. If given the same long as an input, a collision would occur, but it would require that the notebooks are generated at the same millisecond and that they would both get the same result from Random.nextInt(),  so chances are astronomical.

zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/NoteManager.java
Laukkala: Fixed in commit [40d4d47](https://github.com/teragrep/zep_01/pull/328/commits/40d4d479a70aa728552f8ac957495268533ffe6c)

zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/NoteNodeStub.java
```diff
@@ -0,0 +1,61 @@
+package com.teragrep.zep_01.notebook;
+
+import java.io.IOException;
+
+public final class NoteNodeStub implements NoteNode {
+
+    @Override
+    public Note getNote() throws IOException {
+        throw new RuntimeException("NoteNode is a stub!");
```
eemhu: UnsupportedOperationException

Laukkala: Changed in commit [64f75a6](https://github.com/teragrep/zep_01/pull/328/commits/64f75a64261d83d72986683ed413d3769e58a105)

eemhu (approved):

Issue #295, opened 2026-02-20
Create an object for ParagraphConfig's ConfigMap field with type safe methods


Laukkala:
ParagraphConfig contains a configMap field, the contents of which must be inspected with instanceof().
Turn this into a separate object with type safe methods so we can get rid of this kind of introspection.

_Originally posted by @kortemik in https://github.com/teragrep/zep_01/pull/289#discussion_r2815602187_

PR #329, opened 2026-06-09
Issue 103 spark preloader

### LINKED ISSUE #103: Spark preloader server-side implementation
If a notebook contains a Paragraph using Spark interpreter, the interpreter should be started when the note is opened.
The only exception to this is if the notebook contains a Spark configuration paragraph. In that case, the interpreter should remain closed until the paragraphs are run to ensure that  configuration gets applied properly.

- [x] Preloader should run automatically when opening a note
- [x] Preloader should report back when process is complete
- [x] Preloader shouldn't be run when there's a configuration interpreter present within the notebook
- [x] Preloader should be able to preload a subset of interpreters
- [x] Configuration should be done via Interpreter properties
- [x] Handle user-scoped interpreters
- [x] Handle note-scoped interpreters
- [x] Should have logic to void multiple calls to intepreter.open() at the same time when multiple users are present at the same time 
- [x] Should be run in parallel so as to not lock the paragraph while interpreters are loading
- [x] Unit tests present

Some interpreters, such as Spark, use a shared process, but report whether they have been "opened" separately.
Interpreters are identified by their ID's which can include user-supplied strings. This means that Interpreters must be identified exactly. 
Additional information in internal page, search for "Spark Preloader"

Laukkala:
## Description

This PR includes a new Interpreter REST API endpoint for opening an Interpreter.
Sending a request with a defined noteId within its body to api/interpreter/setting/open/{interpreter_setting_name} will open the default interpreter of {interpreter_setting_name} for the logged in user within the notebook matching with the given noteId.

This replaces the preloader implemented in AJS_01, and calls to this endpoint should be implemented on the UI side. Related AJS_01 issue: [#295](https://github.com/teragrep/ajs_01/issues/295)

closes #103 

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality


zeppelin-server/src/test/java/com/teragrep/zep_01/rest/fakes/OpenableInterpreterFake.java
```diff
@@ -0,0 +1,53 @@
...
+import com.teragrep.zep_01.interpreter.InterpreterContext;
+import com.teragrep.zep_01.interpreter.InterpreterException;
+import com.teragrep.zep_01.interpreter.InterpreterResult;
+
+import java.util.Properties;
+
+/**
+ * Fake interpreter that can report whether it has been opened
+ */
+public final class OpenableInterpreterFake extends Interpreter {
```
Nefarious46, line 13: Which class is this fake of? Could not find object called "OpenableInterpreter" making hard to confirm correct expected behavior.

zeppelin-server/src/test/java/com/teragrep/zep_01/rest/InterpreterRestApiTest.java
```diff
@@ -70,6 +72,7 @@ public void setUp() {
     anonymous = new AuthenticationInfo("anonymous");
   }
 
+  @Ignore(value="Flaky test, shuts down JVM")
```
Nefarious46, line 76: Will there be a separate issue to fix/remove these tests if obsolete or broken?

zeppelin-server/src/test/java/com/teragrep/zep_01/rest/fakes/OpenableInterpreterFake.java
Nefarious46: Nevermind its extending instead of implementing.

zeppelin-server/src/test/java/com/teragrep/zep_01/rest/InterpreterRestApiTest.java
```diff
@@ -313,6 +322,80 @@ public void testRestartInterpreterPerNote() throws IOException, InterruptedExcep
...
+    CloseableHttpResponse put = Assertions.assertDoesNotThrow(()-> httpPut("/interpreter/setting/open/"+setting.getId(), jsonRequest));
+
+    // Should result in an 404 error
+    Assertions.assertEquals(404,put.getStatusLine().getStatusCode());
+    Assertions.assertDoesNotThrow(()->put.close());
+
+    // Interpreter should not be opened
+    Assertions.assertFalse(fakeInterpreter.isOpened());
+  }
+
```
Nefarious46, line 440: > The only exception to this is if the notebook contains a Spark configuration paragraph. In that case, the interpreter should remain closed until the paragraphs are run to ensure that configuration gets applied properly.

Should this also be tested?

Nefarious46 (commented):
Sound feature and code. Still needs test for when Note contains Spark configuration to make sure issue is fulfilled.

zeppelin-server/src/test/java/com/teragrep/zep_01/rest/InterpreterRestApiTest.java
Laukkala: I initially thought that UI would be responsible for detecting when it is appropriate to open an Interpreter and would take the presence of configuration paragraphs into account before sending an API call, but I suppose the current implementation is also automatic, so I made opening an Interpreter in a notebook with ConfInterpreters into an error case in commit [4a8afef](https://github.com/teragrep/zep_01/pull/329/commits/4a8afefb95bdca8b6b5c9647f0622ee984c401d1)

I also added some tests and changed the response messages to be more consistent so that UI doesn't have to receive different kinds of messages in case the API call is unsuccessful or not

Nefarious46: Yes the AJS issue is a bit ambiguous regarding where the loader should be called. Perhaps this should be specified more clearly who has the responsibility of acknowledging the need for preloading the interpreter.

PR #333, opened 2026-07-24
Issue 330 version histroy cleared on notebook rename

### LINKED ISSUE #330: Version history is cleared when a notebook is renamed
**Describe the bug**
If a Notebook contains saved versions, renaming said notebook causes versions to be cleared.

**Expected behavior**
Saved versions should be kept intact

**How to reproduce**
Create a new notebook, make some edits
Save the notebook via version control so that you have one or more versions
Rename notebook
Versions gone

**Software version**
4.0.0

Laukkala:
## Description

This PR fixes an issue where version history does not persist through renaming of a notebook by swapping usage of git.log() in GitNotebookRepo to using a RevWalk with a FollowFilter, which does keep track of renaming.

fixes #330 
#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

### DIFF
--- a/zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/repo/Revision.java
+++ b/zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/repo/Revision.java
@@ -8,11 +8,14 @@ public final class Revision {
 /**
  * Represents the 'Revision' a point in life of the notebook
  */
-    public static final Revision EMPTY = new Revision(StringUtils.EMPTY, StringUtils.EMPTY, 0);
 
     public final String id;
     public final String message;
     public final int time;
+
+    public Revision() {
+        this(StringUtils.EMPTY, StringUtils.EMPTY, 0);
+    }
 
     public Revision(final String revId, final String message, final int time) {
         this.id = revId;
@@ -21,7 +24,7 @@ public final class Revision {
     }
 
     public boolean isEmpty() {
-        return this.equals(EMPTY);
+        return (id.equals(StringUtils.EMPTY) && message.equals(StringUtils.EMPTY) && time == 0);
     }
 
     @Override
--- a/zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/repo/Revision.java
+++ b/zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/repo/Revision.java
@@ -24,7 +24,7 @@ public final class Revision {
     }
 
     public boolean isEmpty() {
-        return (id.equals(StringUtils.EMPTY) && message.equals(StringUtils.EMPTY) && time == 0);
+        return (id.isEmpty() && message.isEmpty() && time == 0);
     }
 
     @Override
--- a/zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/repo/Revision.java
+++ b/zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/repo/Revision.java
@@ -1,0 +1,39 @@
+package com.teragrep.zep_01.notebook.repo;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Objects;
+
+public final class Revision {
+/**
+ * Represents the 'Revision' a point in life of the notebook
+ */
+    public static final Revision EMPTY = new Revision(StringUtils.EMPTY, StringUtils.EMPTY, 0);
+
+    public final String id;
+    public final String message;
+    public final int time;
+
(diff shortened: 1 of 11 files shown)


zeppelin-zengine/src/test/java/com/teragrep/zep_01/notebook/repo/GitNotebookRepoTest.java
```diff
@@ -220,6 +221,58 @@ public void addCheckpointTest() throws IOException, GitAPIException {
...
+    // notebook should contain two revisions
+    final List<Revision> revisions = Assertions.assertDoesNotThrow(()->notebookRepo.revisionHistory(TEST_NOTE_ID, TEST_NOTE_PATH, null));
+    Assertions.assertEquals(2,revisions.size());
+
+    // rename notebook
+    final String renamedNotePath = "/my-project/my-renamed-note";
+    Assertions.assertDoesNotThrow(()->notebookRepo.move(TEST_NOTE_ID, TEST_NOTE_PATH,renamedNotePath,null));
+
+    // renamed notebook should have identical commits with the original
+    final List<Revision> renamedRevisions = Assertions.assertDoesNotThrow(()->notebookRepo.revisionHistory(TEST_NOTE_ID, renamedNotePath, null));
```
elliVM, line 251: If we could make `Revision` implement `equals()` and `hashCode()` we could simply compare the lists here instead of looping and also comparing all the class member fields instead of just the id field. 

Looks like the `Revision.IsEmpty()` method uses `equals()`without implementing it, adding `equals()` would fix it also.

elliVM (commented):
Looks good, please see my suggestion before approval

elliVM (dismissed):

eemhu (commented):
LGTM except maybe the revision class could be separated into its own file?

Laukkala:
> LGTM except maybe the revision class could be separated into its own file?

Separated Revision to its own class, and removed static elements from it in commits [24d5eae](https://github.com/teragrep/zep_01/pull/333/commits/24d5eae9f6ecc815b99ce0d1a7c859574e9a1ba3), [eab46fd](https://github.com/teragrep/zep_01/pull/333/commits/eab46fd0a563cfc98134c90ff53659126073b391) and [286c3be](https://github.com/teragrep/zep_01/pull/333/commits/286c3be9b70b0e077756464ec279a496dd8204a7)

zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/repo/Revision.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public Revision(final String revId, final String message, final int time) {
+        this.id = revId;
+        this.message = message;
+        this.time = time;
+    }
+
+    public boolean isEmpty() {
+        return (id.equals(StringUtils.EMPTY) && message.equals(StringUtils.EMPTY) && time == 0);
```
eemhu, line 27: StringUtils.EMPTY could be replaced with "".

eemhu: or just id.isEmpty && message.isEmpty

Laukkala: Fixed in commit [c65e810](https://github.com/teragrep/zep_01/pull/333/commits/c65e810206893d9be9aa09d0261cbcb933a525d1)

eemhu (approved):

zeppelin-zengine/src/main/java/com/teragrep/zep_01/notebook/repo/Revision.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Objects;
+
+public final class Revision {
+/**
+ * Represents the 'Revision' a point in life of the notebook
+ */
+
+    public final String id;
```
eemhu: just noticed that these are public final, is there something preventing them to be private?

Laukkala: These were accessed directly in a couple of tests, but they can be changed to be accessible via methods easily enough. Fixed in commit [a5fab53](https://github.com/teragrep/zep_01/pull/333/commits/a5fab531783890df07e9b6a913447c41f1c8dea0)

eemhu (approved):

PR #335, opened 2026-07-29
Issue 323 interpolation cannot access angularobjects

### LINKED ISSUE #323: DPL interpreter cannot access AngularObjects created via Angular forms
**Describe the bug**
Setting a value in one paragraph using Angular interpreter creates an AngularObject, but that AngularObject is not available for use in other paragraphs using different interpreters, for example DPLInterpreter, even if the AngularObject is set as global.

Same applies to ResourcePools when using z.put() and z.get()

**Expected behavior**
It should be possible to set a value using Angular Interpreter for use in other interpreters

**How to reproduce**

Create paragraph 1 (remember to replace paragraphIds to match your version):
```
%angular
<form>
    <label class="form-label">
        Select column
    </label>
    <div class="input-group">
        <input ng-model="angularValue" type="text" class="form-control" placeholder="Write something" />
        <button ng-click="z.angularUnbind('angularValue'); type="submit" class="btn btn-secondary">
            Reset
        </button>
        <button ng-click="z.angularBind('angularValue', angularValue, 'paragraph_1731580770564_1960409457');" type="submit" class="btn btn-primary">
            Send
        </button>
    </div>
</form>
```
Create paragraph 2:
```
%dpl
index=crud earliest=-5y
| spath
| rename count as operationCount
| stats sum(operationCount) by ${angularValue}

```
Setting a value in paragraph 1 creates an AngularObject, but that AngularObject is not available for use in Paragraph 2

**Software version**
4.0.0

Laukkala:
fixes #323 
## Description
This PR exposes AngularObjects for use in interpolation. Previously only values in ResourcePool could be used. This allows DPL interpreter to access values set via AngularBind methods. These can be used for example to define which columns are used in other DPL queries.

If there are multiple definitions of a variable using the same name, priority will be given to ResourcePool variables, then to AngularObjects by the most specific scope available. (1. paragraph 2. notebook 3. global)

AngularObjects are Interpreter-specific, so if multiple users are sharing one instance of an Interpreter, they will also share AngularObjects for that interpreter, which might cause confusion at best or query injection attacks at worst. By default though, each user has their own DPL interpreter and their own AngularObjects, but it should be considered that sharing an Interpreter session also shares access to AngularObjects, which can be arbitrary strings.

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality


zeppelin-interpreter/src/main/java/com/teragrep/zep_01/interpreter/AbstractInterpreter.java
```diff
@@ -42,16 +47,45 @@ public InterpreterResult interpret(String st,
...
+      for (AngularObject angularObject : allObjects) {
+        replacementMap.put(angularObject.getName(),angularObject.get());
+      }
+    }
+
+    // StringSubstitutor wants Map so convert ResourcePool if exists
+    if(resourcePool != null){
+      final ResourceSet resources = resourcePool.getAll();
+      for (int i = 0; i < resources.size(); i++) {
+        String name = resources.get(i).getResourceId().getName();
```
elliVM: Local variables `name` and `value` could be final

zeppelin-interpreter/src/main/java/com/teragrep/zep_01/interpreter/AbstractInterpreter.java
```diff
@@ -42,16 +47,45 @@ public InterpreterResult interpret(String st,
...
+
+      // Add each object to a Map from the List. Order of the List matters, value of a duplicated key in the Map will be the last occurrence of that key in the List.
+      for (AngularObject angularObject : allObjects) {
+        replacementMap.put(angularObject.getName(),angularObject.get());
+      }
+    }
+
+    // StringSubstitutor wants Map so convert ResourcePool if exists
+    if(resourcePool != null){
+      final ResourceSet resources = resourcePool.getAll();
```
elliVM, line 80: Since `ResourceSet` extends `LinkedList`,  it would be better to iterate using an enhanced for-loop rather than indexing with `get(i)`? We would avoid a repeated traversal of the linked list
```
for (Resource resource : resources) {
    replacementMap.put(resource.getResourceId().getName(), resource.get());
}
``` 
would this work?

elliVM (changes requested):

zeppelin-interpreter/src/main/java/com/teragrep/zep_01/interpreter/AbstractInterpreter.java
Laukkala: Fixed in commit [f95174b](https://github.com/teragrep/zep_01/pull/335/commits/f95174b72977bc27bc57be90554b808da42d631f)

zeppelin-interpreter/src/main/java/com/teragrep/zep_01/interpreter/AbstractInterpreter.java
Laukkala: Fixed in commit [f95174b](https://github.com/teragrep/zep_01/pull/335/commits/f95174b72977bc27bc57be90554b808da42d631f)

elliVM (approved):
LGTM!

eemhu (approved):

